PR #1, opened 2020-06-10
Update README.md

mkitez:


DirectProgramming/DPC++/DenseLinearAlgebra/simple-add/sample.json
```diff
@@ -6,5 +6,74 @@
   "languages": [{"cpp": {"properties": {"projectOptions": [{"projectType": "makefile"}]}}}],
   "targetDevice": ["CPU", "GPU", "FPGA"],
   "os": ["linux", "windows"],
-  "builder": ["ide", "make"]
+  "builder": ["ide", "make"],
+  "configurations": {
```
xmnboy, line 10: I think the name of the CI testing object should be more direct. The word `configurations` does not communicate to me that this contains CI tests. How about calling the object `ciTests`?
```suggestion
  "ciTests": {
```

DirectProgramming/DPC++/DenseLinearAlgebra/simple-add/sample.json
```diff
@@ -6,5 +6,74 @@
...
+        "build": [
+          "make all"
+        ],
+        "run": [
+          "make run"
+        ],
+        "clean": [
+          "make clean"
+        ]
+      },
```
xmnboy, line 23: Roman -- I think my suggestion (below) is consistent with what you were alluding to, yesterday. Rather than being specific to the build, you might want to have a more general set of "phases" that direct the tests to be performed (mind you, I'm _not_ a CI expert, so take my suggestions with a "grain of salt.").

Perhaps something like the following, which are overly complex to make a point, not to indicate that all tests would be like this (and where all array element "actions" are assumed to start from the root of the sample):
```JSON
    "linux": [
      {
        "id": "cpu_usm",
        "setup": [ 
          [ "mkdir -p test/test && cd test/test && echo 'x=y' >test.txt", "/success-regex/", "/fail-regex/" ],
          [ "export TEST=test ; touch abc ; my-setup-script.sh || true", "/success-regex/", "/fail-regex/" ]
        ],
        "test": [
          [ "mkdir build && cd build && cmake ..", "/success-regex/", "/fail-regex/" ]
        ],
        "clean": [
          [ "make clean", "", "" ]
        ]
      }
    ],
```
Where the idea is that a sequence of commands that could be executed at the shell, with optional "success" and "fail" regex strings provided to spot a success or failure based on the output (if any) of the commands. I've made the above overly complex just to make the point that I don't think you want to assume a particular build tool, rather that you just want to know "what do I type to setup for a test" and "what do I type to test" and "what key phrases/output" do I expect to see on success/failure?

I would also assume that you would continue to check the return code from each "action" as part of the testing, assuming that a non-zero value is a pass and anything else is a fail...

Anyway, just a thought, any further than what I have suggested above you'd probably want to use something like YAML rather than JSON to use as the testing template. 

Again, I emphasize that this is just an idea. :)

mkitez: What goes into the test command line is completely up to the sample owner. If they want to validate output against a regex, they can do it as long as the command returns zero on success and non-zero otherwise. I don't think we can reliably tell in CI whether the sample is running well other than by checking the exit status.

I'm still not sure how to break down the test commands, by build/run/clean stages or by test phases as you suggested or have them as a single command at all. Or there can be arbitrary stages. For example some samples may omit the `clean` step, others may include more steps, e.g. `pre-build` but I'm worried it may complicate the things.

PR #4, opened 2020-06-29
Create CODEOWNERS

anjgola:


CODEOWNERS
```diff
@@ -0,0 +1,24 @@
...
+# subdirectories.
+/DirectProgramming/DPC++/ @JoeOster @moushumi-maria, @pmpeter1
+/DirectProgramming/Jupyter/ @JoeOster @praveenkk123
+/DirectProgramming/C++/ @JoeOster @pmpeter1 @Propanu
+/DirectProgramming/Fortran/ @JoeOster @pmpeter1 @Propanu
+/Tools/ @JoeOster @Propanu
+/Libraries/ @JoeOster @Propanu @mav-intel @JoeOster
+
+# Reviwers for all sample.json modification 
+*sample.json @pfische1
```
mkitez, line 20: Can you please add myself here as well? I would like to be aware of the `ciTests` field changes in particular

anjgola: Ok

anjgola:
@mkitez  I can use your help regarding this PR. I haven't merged this because it is not a good solution and I feel stuck on how to proceed.
Github codeowners conditions are not aggregated and only the last condition met is applied. That means if someone edits a sample.json and a README.md file, then whatever condition I list at the end of this file will be executed. This doesn't help.
I see this article that solves a similar problem : https://bionic.fullstory.com/taming-github-codeowners-with-bots/
Do you think we need to do something like this?


In short term, we could just go with 3-4 blanket reviewers for all files for now but we need a better solution.

mkitez:
@anjgola sounds good. Let's go with what you suggested for now, we can look into the approach described in the article moving forward.

PR #24, opened 2020-07-13
Update README.md

JoeOster:


anjgola:
Can we put this as third-party-programs.txt file instead of README. We need to put other landing page information in README.

common/dpc_common.hpp
```diff
@@ -0,0 +1,52 @@
+//==============================================================
```
mkitez, line 1: Why are we committing this file here? Isn't it supposed to be inside oneAPI package only?

README.md
```diff
@@ -1,2 +1,2 @@
-# oneAPI-samples
+oneAPI-samples
```
mkitez, line 1: Looks like you are changing header to a paragraph, is this intentional?

common/dpc_common.hpp
JoeOster: per request by Swapna, it is for reference only.

README.md
JoeOster: This is just a place holder, we are working on a readme, so do not approve the merge as of yet

PR #36, opened 2020-07-17
Update sample.json

JoeOster:


anjgola:
Weird..why are windows and linux tests status not being reported. @mkitez  Please check when you can.

JoeOster:
@anjgola I think the difference is that I made a branch off master and most people are forking and pulling a fork into master.

PR #39, opened 2020-07-20
Fixed readme and VS configuration

varsha-madananth:


anjgola:
What are these .suo and .db file. Do they need to be merged?

PR #48, opened 2020-07-28
Clang-format check deleted

andrey4latyshev:
Clang-format check will be moved to CI repository


anjgola:
This looks good, I think it will work.

PR #59, opened 2020-08-07
WIP Adding Nbody sample code to the repo

mpanoop:
Adding Nbody sample code to the repo


DirectProgramming/DPC++/N-BodyMethods/Nbody/sample.json
```diff
@@ -0,0 +1,29 @@
+{
+  "guid": "D5BFD7D5-737E-4CC9-84EE-FE94339B3DCA"
```
anjgola, line 2: Incorrect Syntax, comma is missing, sample will not pass CI. Please run jsonlint on this file before submitting merge request.

pmpeter1:
Many of the variable names  don't follow the consistent naming style used in the samples.  totTime , distanceSqr , distanceInv , elapsedSeconds.

In general the code is only minimally commented, and would benefit from more explainations on why functions exist, and on the phases of the algorithms.

Check all of your "new" calls to make sure you have matching deletes to avoid memory leaks.   Often using std::vector is preferred to make this automatic to avoid these problems.  

The dpc_common include should include the recommended comment to show where the user can find this shared file.

You should not be passing const int by reference, it should just be by value.  const int &N  --> const int N.       Obejcts benefit from being passed by const ref to avoid creating a new copy of the object.  This is not true for scalars.

JoeOster:
@mpanoop - Can you look at this and provide a status for the requested changes

PR #61, opened 2020-08-07
Add pull request template

anjgola:


anjgola:
I made slight modifications to accommodate your changes, I think this is the best we can do for now.

anjgola:
@mkitez  Any thoughts on why linux CI is failing. seems like the job is looking for report results to post here and when it does not find them it errs out?

PR #65, opened 2020-08-10
Fix STB includes for protex - C++/MandelbrotOMP

ethanhirsch:
Changes in DirectProgramming/C++/MandelbrotOMP


sravanikonda:
LGTM
@mkitez Any idea why the Linux tests were not executed?

PR #163, opened 2020-09-17
Hidden Markov Model Sample: update sample.json

andrey4latyshev:
# Description

Change of the sample's category in sample.json file

The PR is waiting for Swapna's approval so it is tagged as 'WIP'


anjgola (commented):
Why make change to the base toolkit name?

andrey4latyshev:
> Why make change to the base toolkit name?

I made this according to this page:
https://github.com/oneapi-src/oneAPI-samples/wiki/sample-json-categories-array

anjgola:
Can you elaborate more what to infer from the link you provided and how it relates to the change you are making here.

The problem in the email is described as Having wrong category. I see in the screenshot that for the hidden markov sample the sub-sub-category 'oneAPI DPC++ compiler' is unlike the other majority sample so I infer from that the compiler part should change to become 'Intel oneAPI DPC++/C++ Compiler/CPU and GPU. 
But I see from the screenshot that the toolkit part is right where it should be. It is currently under Intel oneAPI Base Toolkit and that is where it is expected to be. So I don't think this PR should be changing the toolkit part. 

Adding @xmnboy  for review.

anjgola:
please use the suggestion : “Intel® oneAPI Base Toolkit/oneAPI DPC++/C++ Compiler/CPU and GPU”

anjgola (approved):

PR #167, opened 2020-09-18
Add compute_units DPC++ FPGA tutorial

jessicadavies-intel:
#  Description

This tutorial shows how to make multiple copies of kernels, called compute_units.

# Type of Change

# How Has This Been Tested?

- [X] Command Line

### DIFF
--- a/DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/CMakeLists.txt
+++ b/DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/CMakeLists.txt
@@ -1,0 +1,11 @@
+set(CMAKE_CXX_COMPILER "dpcpp")
+
+cmake_minimum_required (VERSION 2.8)
+
+project(ComputeUnits)
+
+set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR})
+set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR})
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR})
+
+add_subdirectory (src)
--- a/DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/README.md
+++ b/DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/README.md
@@ -42,7 +42,7 @@ The `Source` kernel reads data from host memory and writes it to the first pipe:
 void SourceKernel(queue &q, float data) {
 
   q.submit([&](handler &h) {
-    h.single_task<Source>([=]() { Pipes::PipeAt<0>::write(data); });
+    h.single_task<Source>([=] { Pipes::PipeAt<0>::write(data); });
   });
 }
 ```
@@ -50,14 +50,17 @@ void SourceKernel(queue &q, float data) {
 At the end of the chain, the `Sink` kernel reads data from the last pipe and returns it to the host:
 
 ``` c++
-void SinkKernel(queue &q, std::array<float, 1> &out_data) {
+void SinkKernel(queue &q, float *out_data) {
 
-  buffer<float, 1> out_buf(out_data.data(), 1);
+  // The verbose buffer syntax is necessary here,
+  // since out_data is just a single scalar value
+  // and its size can not be inferred automatically
+  buffer<float, 1> out_buf(out_data, 1);
 
   q.submit([&](handler &h) {
     auto out_accessor = out_buf.get_access<access::mode::write>(h);
     h.single_task<Sink>(
-        [=]() { out_accessor[0] = Pipes::PipeAt<kEngines>::read(); });
+        [=] { out_accessor[0] = Pipes::PipeAt<kEngines>::read(); });
   });
 }
 ```
@@ -66,7 +69,7 @@ void SinkKernel(queue &q, std::array<float, 1> &out_data) {
 
 ``` c++
 intelfpga::submit_compute_units<kEngines, ChainComputeUnit>(q, [=](auto ID) {
-    float f = Pipes::PipeAt<ID>::read();
+    auto f = Pipes::PipeAt<ID>::read();
     Pipes::PipeAt<ID + 1>::write(f);
   });
 ```
--- a/DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/README.md
+++ b/DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/README.md
@@ -1,0 +1,151 @@
+
+# Using Compute Units To Duplicate Kernels
+This FPGA tutorial showcases a design pattern that allows you to make multiple copies of a kernel, called compute units.
+
+***Documentation***: The [oneAPI DPC++ FPGA Optimization Guide](https://software.intel.com/content/www/us/en/develop/documentation/oneapi-fpga-optimization-guide)  provides comprehensive instructions for targeting FPGAs through DPC++. The [oneAPI Programming Guide](https://software.intel.com/en-us/oneapi-programming-guide) is a general resource for target-independent DPC++ programming. 
+
+| Optimized for                     | Description
+---                                 |---
+| OS                                | Linux* Ubuntu* 18.04; Windows* 10
+| Hardware                          | Intel® Programmable Acceleration Card (PAC) with Intel Arria® 10 GX FPGA; <br> Intel® Programmable Acceleration Card (PAC) with Intel Stratix® 10 SX FPGA
+| Software                          | Intel® oneAPI DPC++ Compiler (Beta) <br> Intel® FPGA Add-On for oneAPI Base Toolkit 
+| What you will learn               | A design pattern to generate multiple compute units in DPC++ using template metaprogramming
+| Time to complete                  | 15 minutes
+
+_Notice: Limited support in Windows*; compiling for FPGA hardware is not supported in Windows*_
+
+## Purpose
+The performance of a design can sometimes by increased by making multiple copies of kernels, to make full use of the FPGA resources. Each kernel copy is called a compute unit.  
+
+This tutorial provides a header file that defines an abstraction for making multiple copies of a single-task kernel. This header can be used in any DPC++ design and can be extended as necessary.
+
+### Compute units example 
+
+This code sample builds on the pipe_array tutorial. It includes the header files `pipe_array.hpp`, `pipe_array_internal.hpp`, and `unroller.hpp` from the pipe_array tutorial, and provides the `compute_units.hpp` header as well.
+
+This code sample defines a `Source` kernel, and a `Sink` kernel. Data is read from host memory by the `Source` kernel, and sent to the `Sink` kernel via a chain of compute units. The number of compute units in the chain between `Source` and `Sink` is determined by the following parameter:
+
+``` c++
+constexpr size_t kEngines = 5;
+```
+
+Data is passed between kernels using pipes. The number of pipes needed is equal to the length of the chain, i.e., the number of compute units, plus one.
+This array of pipes is declared as follows:
+
+``` c++
+using Pipes = PipeArray<class MyPipe, float, 1, kEngines + 1>;
+```
+
+The `Source` kernel reads data from host memory and writes it to the first pipe:
+
+``` c++
+void SourceKernel(queue &q, float data) {
+
+  q.submit([&](handler &h) {
+    h.single_task<Source>([=]() { Pipes::PipeAt<0>::write(data); });
+  });
+}
+```
+
+At the end of the chain, the `Sink` kernel reads data from the last pipe and returns it to the host:
+
+``` c++
+void SinkKernel(queue &q, std::array<float, 1> &out_data) {
+
+  buffer<float, 1> out_buf(out_data.data(), 1);
+
+  q.submit([&](handler &h) {
+    auto out_accessor = out_buf.get_access<access::mode::write>(h);
+    h.single_task<Sink>(
+        [=]() { out_accessor[0] = Pipes::PipeAt<kEngines>::read(); });
+  });
+}
+```
+
+ The following line defines the functionality of each compute unit, and submits each compute unit to the given queue `q`. The number of copies to submit is provided by the first template parameter of `submit_compute_units`, while the second template parameter allows you to give the compute units a shared base name, e.g., `ChainComputeUnit`. The functionality of each compute unit must be specified by a generic lambda expression, that takes a single argument (with type `auto`) that provides an ID for each compute unit:
+
+``` c++
+intelfpga::submit_compute_units<kEngines, ChainComputeUnit>(q, [=](auto ID) {
+    float f = Pipes::PipeAt<ID>::read();
+    Pipes::PipeAt<ID + 1>::write(f);
+  });
+```
+
+Each compute unit in the chain from `Source` to `Sink` must read from a unique pipe, and write to the next pipe. As seen above, each compute unit knows its own ID and therefore its behavior can depend on this ID. Each compute unit in the chain will read from pipe `ID` and write to pipe `ID + 1`.
+
+## Key Concepts
+* A design pattern to generate multiple compute units in DPC++ using template metaprogramming
+
+## License  
+This code sample is licensed under MIT license.
+
+
+## Building the `compute_units` Tutorial
+
+### Include Files
+The included header `dpc_common.hpp` is located at `%ONEAPI_ROOT%\dev-utilities\latest\include` on your development system.
+
+### Running Samples in DevCloud
+If running a sample in the Intel DevCloud, remember that you must specify the compute node (fpga_compile or fpga_runtime) as well as whether to run in batch or interactive mode. For more information see the Intel® oneAPI Base Toolkit Get Started Guide ([https://devcloud.intel.com/oneapi/get-started/base-toolkit/](https://devcloud.intel.com/oneapi/get-started/base-toolkit/)).
+
+When compiling for FPGA hardware, it is recommended to increase the job timeout to 12h.
+
+### On a Linux* System
+
+1. Generate the `Makefile` by running `cmake`.
+     ```
+   mkdir build
+   cd build
+   ```
+   To compile for the Intel® PAC with Intel Arria® 10 GX FPGA, run `cmake` using the command:  
+    ```
+    cmake ..
+   ```
+   Alternatively, to compile for the Intel® PAC with Intel Stratix® 10 SX FPGA, run `cmake` using the command:
+
+   ```
+   cmake .. -DFPGA_BOARD=intel_s10sx_pac:pac_s10
+   ```
+
+2. Compile the design through the generated `Makefile`. The following build targets are provided, matching the recommended development flow:
+
+   * Compile for emulation (fast compile time, targets emulated FPGA device): 
+      ```
+      make fpga_emu
+      ```
+   * Generate the optimization report: 
+     ```
+     make report
+     ``` 
+   * Compile for FPGA hardware (longer compile time, targets FPGA device): 
+     ```
+     make fpga
+     ``` 
+3. (Optional) As the above hardware compile may take several hours to complete, an Intel® PAC with Intel Arria® 10 GX FPGA precompiled binary can be downloaded <a href="https://software.intel.com/content/dam/develop/external/us/en/documents/compute_units.fpga.tar.gz" download>here</a>.
+
+ ### In Third-Party Integrated Development Environments (IDEs)
+
+You can compile and run this tutorial in the Eclipse* IDE (in Linux*). For instructions, refer to the following link: [Intel® oneAPI DPC++ FPGA Workflows on Third-Party IDEs](https://software.intel.com/en-us/articles/intel-oneapi-dpcpp-fpga-workflow-on-ide)
+
+
+## Examining the Reports
+Locate `report.html` in the `compute_units_report.prj/reports/` or `compute_units_s10_pac_report.prj/reports/` directory. Open the report in any of Chrome*, Firefox*, Edge*, or Internet Explorer*.
+
+You can visualize the kernels and pipes generated by looking at the "System Viewer" section of the report. Note that each compute unit is shown as a unique kernel in the reports, with names `ChainComputeUnit<0>`, `ChainComputeUnit<1>`, etc.
+
+## Running the Sample
+
+ 1. Run the sample on the FPGA emulator (the kernels execute on the CPU):
+     ```
+     ./compute_units.fpga_emu     (Linux)
+     compute_units.fpga_emu.exe   (Windows)
+     ```
+2. Run the sample on the FPGA device:
+     ```
+     ./compute_units.fpga         (Linux)
+     ```
+
+### Example of Output
+```
+PASSED: The results are correct
+```
--- a/DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/build.ninja
+++ b/DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/build.ninja
@@ -5,8 +5,8 @@ emulator_target = ${target_name}.fpga_emu.exe
 report_target = ${target_name}_report.a
 report_target_s10_pac = ${target_name}_s10_pac_report.a
 
-hardware_flags = -fintelfpga -Xshardware -std=c++14
-emulator_flags = -fintelfpga -DFPGA_EMULATOR -std=c++14
+hardware_flags = -fintelfpga -Xshardware
+emulator_flags = -fintelfpga -DFPGA_EMULATOR
 
 rule build_fpga_emu
   command = dpcpp /GX ${emulator_flags} $in -o $out
--- a/DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/build.ninja
+++ b/DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/build.ninja
@@ -1,0 +1,30 @@
+source_file = compute_units.cpp
+target_name = compute_units
+
+emulator_target = ${target_name}.fpga_emu.exe
+report_target = ${target_name}_report.a
+report_target_s10_pac = ${target_name}_s10_pac_report.a
+
+hardware_flags = -fintelfpga -Xshardware -std=c++14
+emulator_flags = -fintelfpga -DFPGA_EMULATOR -std=c++14
+
+rule build_fpga_emu
+  command = dpcpp /GX ${emulator_flags} $in -o $out
+
+rule gen_report
+  command = dpcpp /GX ${hardware_flags} -Xsboard=intel_a10gx_pac:pac_a10 -fsycl-link $in -o $out
+
+rule gen_report_s10_pac
+  command = dpcpp /GX ${hardware_flags} -Xsboard=intel_s10sx_pac:pac_s10 -fsycl-link $in -o $out
+
+# FPGA emulator
+build fpga_emu: phony ${emulator_target}
+build ${emulator_target}: build_fpga_emu ${source_file}
+
+# report
+build report: phony ${report_target}
+build ${report_target}: gen_report ${source_file}
+
+# report (S10 PAC)
+build report_s10_pac: phony ${report_target_s10_pac}
+build ${report_target_s10_pac}: gen_report_s10_pac ${source_file}
--- a/DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/compute_units.cpp
+++ b/DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/compute_units.cpp
@@ -45,11 +45,12 @@ void SourceKernel(queue &q, float data) {
 }
 
 // Get the data out of the chain and return it to the host
-void SinkKernel(queue &q, std::array<float, 1> &out_data) {
+void SinkKernel(queue &q, float *out_data) {
 
-  // Use verbose SYCL 1.2 syntax for the output buffer.
-  // (This will become unnecessary in a future compiler version.)
-  buffer<float, 1> out_buf(out_data.data(), 1);
+  // The verbose buffer syntax is necessary here,
+  // since out_data is just a single scalar value
+  // and its size can not be inferred automatically
+  buffer<float, 1> out_buf(out_data, 1);
 
   q.submit([&](handler &h) {
     auto out_accessor = out_buf.get_access<access::mode::write>(h);
@@ -66,7 +67,7 @@ int main() {
   INTEL::fpga_selector device_selector;
 #endif
 
-  std::array<float, 1> out_data = {0};
+  float out_data = 0;
 
   try {
     queue q(device_selector, dpc_common::exception_handler);
@@ -77,12 +78,12 @@ int main() {
     // Enqueue the chain of kEngines compute units
     // Compute unit must take a single argument, its ID
     submit_compute_units<kEngines, ChainComputeUnit>(q, [=](auto ID) {
-      float f = Pipes::PipeAt<ID>::read();
+      auto f = Pipes::PipeAt<ID>::read();
       Pipes::PipeAt<ID + 1>::write(f);
     });
 
     // Enqueue the Sink kernel
-    SinkKernel(q, out_data);
+    SinkKernel(q, &out_data);
 
   } catch (sycl::exception const &e) {
     // Catches exceptions in the host code
@@ -99,9 +100,9 @@ int main() {
   }
 
   // Verify result
-  if (out_data[0] != kTestData) {
+  if (out_data != kTestData) {
     std::cout << "FAILED: The results are incorrect\n";
-    std::cout << "Expected: " << kTestData << " Got: " << out_data[0] << "\n";
+    std::cout << "Expected: " << kTestData << " Got: " << out_data << "\n";
     return 1;
   }
 
--- a/DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/compute_units.cpp
+++ b/DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/compute_units.cpp
@@ -5,6 +5,7 @@
 // =============================================================
 #include <CL/sycl.hpp>
 #include <iostream>
+
 // dpc_common.hpp can be found in the dev-utilities include folder,
 // e.g., $ONEAPI_ROOT/dev-utilities/latest/include/dpc_common.hpp
 #include "dpc_common.hpp"
--- a/DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/compute_units.cpp
+++ b/DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/compute_units.cpp
@@ -1,0 +1,109 @@
+//==============================================================
+// Copyright Intel Corporation
+//
+// SPDX-License-Identifier: MIT
+// =============================================================
+#include <CL/sycl.hpp>
+#include <iostream>
+// dpc_common.hpp can be found in the dev-utilities include folder,
+// e.g., $ONEAPI_ROOT/dev-utilities/latest/include/dpc_common.hpp
+#include "dpc_common.hpp"
+#include "compute_units.hpp"
+#include "pipe_array.hpp"
+
+// Header locations and some DPC++ extensions changed between beta09 and beta10
+// Temporarily modify the code sample to accept either version
+#define BETA09 20200827
+#if __SYCL_COMPILER_VERSION <= BETA09
+  #include <CL/sycl/intel/fpga_extensions.hpp>
+  namespace INTEL = sycl::intel;  // Namespace alias for backward compatibility
+#else
+  #include <CL/sycl/INTEL/fpga_extensions.hpp>
+#endif
+
+
+using namespace sycl;
+
+constexpr float kTestData = 555;
+constexpr size_t kEngines = 5;
+
+using Pipes = PipeArray<class MyPipe, float, 1, kEngines + 1>;
+
+// Forward declaration of the kernel name
+// (This will become unnecessary in a future compiler version.)
+class Source;
+class Sink;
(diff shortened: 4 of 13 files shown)


jessicadavies-intel:
@akertesz @AlbyB @pmpeter1 @tomlenth please review.

akertesz:
Your commit includes several superfluous files -- a.out in the src directory, and a bunch of binary files. Please remove these.

akertesz:
@pmpeter1 -- The README contains text indicating where to find the dpc_common.hpp file. You want to see this in the *cpp as well?

pmpeter1:
Yes, the coding guidelines recommend a 1-line comment showing the path to the directory.   The way we are delivering shared headers is a bit awkward so having it in both places is recommended.

JoeOster (approved):

jessicadavies-intel:
@akertesz thank you for pointing me to the unrelated binary files that got deleted accidentally. I used a hard rebase to clean up the branch, which closed this PR automatically.

The new commit includes the dpc_common.h comment requested by @pmpeter1

pmpeter1 (dismissed):

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/compute_units.cpp
```diff
@@ -0,0 +1,109 @@
+//==============================================================
+// Copyright Intel Corporation
+//
+// SPDX-License-Identifier: MIT
+// =============================================================
+#include <CL/sycl.hpp>
+#include <iostream>
+// dpc_common.hpp can be found in the dev-utilities include folder,
```
JoeOster, line 8: add a space here to separate the includes

JoeOster (changes requested):

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/compute_units.cpp
jessicadavies-intel: Ok, I added a blank line before the comment.

JoeOster (approved):

tomlenth:
Readme looks fine.  thank you

akertesz:
Shouldn't merge this until Paul approves, and we sort out why CI isn't running.

pmpeter1 (approved):

JoeOster (changes requested):
Need to figure out ci failure, here are a part of the error logs
Note: We should be targeting c++17
. 
dpcpp /GX -fintelfpga -DFPGA_EMULATOR -std=c++14 compute_units.cpp -o compute_units.fpga_emu.exe

dpcpp: warning: unknown argument ignored in clang-cl '-std=c++14'; did you mean '-Qstd=c++14'? [-Wunknown-argument]

In file included from compute_units.cpp:13:

./pipe_array.hpp(38,19): error: 'INTEL' is not a class, namespace, or enumeration

        cl::sycl::INTEL::pipe<StructId<idxs...>, BaseTy, depth>;

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/README.md
```diff
@@ -0,0 +1,151 @@
...
+| OS                                | Linux* Ubuntu* 18.04; Windows* 10
+| Hardware                          | Intel® Programmable Acceleration Card (PAC) with Intel Arria® 10 GX FPGA; <br> Intel® Programmable Acceleration Card (PAC) with Intel Stratix® 10 SX FPGA
+| Software                          | Intel® oneAPI DPC++ Compiler (Beta) <br> Intel® FPGA Add-On for oneAPI Base Toolkit 
+| What you will learn               | A design pattern to generate multiple compute units in DPC++ using template metaprogramming
+| Time to complete                  | 15 minutes
+
+_Notice: Limited support in Windows*; compiling for FPGA hardware is not supported in Windows*_
+
+## Purpose
+The performance of a design can sometimes by increased by making multiple copies of kernels, to make full use of the FPGA resources. Each kernel copy is called a compute unit.
```
AlbyB, line 18: Though the performance consideration is not listed in "what you will learn"/"key concepts" right now, might be nice to show the performance gain in this tutorial from not using compute units -> using N compute units if applicable (doesn't necessarily need to be in this revision/commit).

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/pipe_array.hpp
```diff
@@ -0,0 +1,43 @@
...
+  template <size_t... idxs>
+  struct VerifyIndices {
+    static_assert(sizeof...(idxs) == sizeof...(dims),
+                  "Indexing into a PipeArray requires as many indices as "
+                  "dimensions of the PipeArray.");
+    static_assert(VerifierDimLayer<dims...>::template VerifierIdxLayer<
+                      idxs...>::IsValid(),
+                  "Index out of bounds");
+    using VerifiedPipe =
+        cl::sycl::INTEL::pipe<StructId<idxs...>, BaseTy, depth>;
```
akertesz, line 38: This is the cause of the compile error. For beta09/10 compatibility, the code aliases the INTEL namespace (new to beta10) back to sycl::intel (from beta09). But here the INTEL namespace is fully scoped.

 Please change it to:
`INTEL::pipe<StructId<idxs...>, BaseTy, depth>;`

This compiles for me on beta09 after this change.

jessicadavies-intel: Thank you for the fix!

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/README.md
jessicadavies-intel: Thank you for the suggestion, I will keep this in mind for future versions of the tutorial.

akertesz: If you make such a change in a future revision, we should apply it across all similar tutorials. This tutorial is structurally similar to the loop_unroll FPGA feature tutorial, which entails a similar area/throughput tradeoff.

AlbyB (dismissed):

akertesz (dismissed):

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/build.ninja
```diff
@@ -0,0 +1,30 @@
+source_file = compute_units.cpp
```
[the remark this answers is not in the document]
keryell: Actually in the documentation on compiling above, you use explicit `make` instead of `cmake --build`, so I cannot see how some Ninja code is getting involved while not using Ninja...

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/compute_units.cpp
```diff
@@ -0,0 +1,110 @@
...
+    h.single_task<Source>([=] { Pipes::PipeAt<0>::write(data); });
+  });
+}
+
+// Get the data out of the chain and return it to the host
+void SinkKernel(queue &q, std::array<float, 1> &out_data) {
+
+  // Use verbose SYCL 1.2 syntax for the output buffer.
+  // (This will become unnecessary in a future compiler version.)
+  buffer<float, 1> out_buf(out_data.data(), 1);
```
keryell, line 52: Probably
```c++
   buffer out_buf(out_data.data(), 1);
```

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/compute_units.cpp
```diff
@@ -0,0 +1,110 @@
...
+  try {
+    queue q(device_selector, dpc_common::exception_handler);
+
+    // Enqueue the Source kernel
+    SourceKernel(q, kTestData);
+
+    // Enqueue the chain of kEngines compute units
+    // Compute unit must take a single argument, its ID
+    submit_compute_units<kEngines, ChainComputeUnit>(q, [=](auto ID) {
+      float f = Pipes::PipeAt<ID>::read();
```
keryell, line 80:
```suggestion
      auto f = Pipes::PipeAt<ID>::read();
```

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/compute_units.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+int main() {
+
+#if defined(FPGA_EMULATOR)
+  INTEL::fpga_emulator_selector device_selector;
+#else
+  INTEL::fpga_selector device_selector;
+#endif
+
+  std::array<float, 1> out_data = {0};
```
keryell, line 69: Why an array of 1 element for a scalar?

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/compute_units.cpp
```diff
@@ -0,0 +1,110 @@
...
+      float f = Pipes::PipeAt<ID>::read();
+      Pipes::PipeAt<ID + 1>::write(f);
+    });
+
+    // Enqueue the Sink kernel
+    SinkKernel(q, out_data);
+
+  } catch (sycl::exception const &e) {
+    // Catches exceptions in the host code
+    std::cout << "Caught a SYCL host exception:\n" << e.what() << "\n";
```
keryell, line 89:
```suggestion
    std::cerr << "Caught a SYCL host exception:\n" << e.what() << "\n";
```

keryell (changes requested):
Nice to have a more complex FPGA example.
Please explain more the code with some comments, since this is a tutorial.
In https://github.com/intel/llvm/pull/2395 I have already reviewed this code with some suggestions about some syntax simplification for your `pipe_array`.
@mkinsner @jbrodman I would love simpler code when this is about FPGA... :-)

keryell:
You could remove a + in the PR title `Add compute_units DPC+++ FPGA tutorial` and also talk about "pipe array"?

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/compute_units.cpp
```diff
@@ -0,0 +1,110 @@
+//==============================================================
+// Copyright Intel Corporation
+//
+// SPDX-License-Identifier: MIT
+// =============================================================
+#include <CL/sycl.hpp>
```
keryell, line 6: Probably in 2020:
```suggestion
#include <SYCL/sycl.hpp>
```

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/compute_units.cpp
akertesz: There was a copy-back bug in an earlier compiler version that prevented the use of the concise syntax. I believe that it is fixed in beta09, but please test functionality in both beta09 and the WIP beta10 compiler if you make this change.

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/build.ninja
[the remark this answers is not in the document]
akertesz: All FPGA samples are temporarily using ninja for Windows build while we await the resolution of a DPC++ Windows compatibility issue with CMake.

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/compute_units.cpp
tyoungsc:
```c++
buffer out_buf(out_data.data(), 1);
```
won't work, but this will
```c++
buffer out_buf(out_data);
```

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/compute_units.cpp
tyoungsc: I think this was likely so we could use the nice buffer constructor:
```c++
buffer out_buf(out_data);
```
Otherwise, we have to use:
```c++
buffer<float,1> out_buf(&out_data, 1);
```

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/compute_units.cpp
tyoungsc: This will break compile

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/compute_units.cpp
akertesz: Yes, that's the preferred syntax. I checked -- the issue that prevented its use is resolved in beta09.

All code samples that implemented the beta08 workaround should be updated to use the concise syntax.

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/compute_units.cpp
jessicadavies-intel: Yes I think we need to keep CL/sycl.hpp, as in: https://github.com/intel/llvm/blob/sycl/sycl/include/CL/sycl.hpp

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/README.md
```diff
@@ -0,0 +1,151 @@
...
+using Pipes = PipeArray<class MyPipe, float, 1, kEngines + 1>;
+```
+
+The `Source` kernel reads data from host memory and writes it to the first pipe:
+
+``` c++
+void SourceKernel(queue &q, float data) {
+
+  q.submit([&](handler &h) {
+    h.single_task<Source>([=]() { Pipes::PipeAt<0>::write(data); });
```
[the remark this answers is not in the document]
jessicadavies-intel: Thank you for pointing out that the README code snippets were out-of-sync with the actual code. This is fixed now.

The pipe_array library was actually copied from an existing tutorial: https://github.com/oneapi-src/oneAPI-samples/tree/master/DirectProgramming/DPC%2B%2BFPGA/Tutorials/DesignPatterns/pipe_array

I plan to make changes to the pipe_array API in both tutorials as a separate change.

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+set(CMAKE_CXX_COMPILER "dpcpp")
+
+cmake_minimum_required (VERSION 2.8)
```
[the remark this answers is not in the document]
jessicadavies-intel: We are testing with cmake 2.8.4 so I believe this is the intended minimum version.

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/README.md
```diff
@@ -0,0 +1,151 @@
...
+  });
+}
+```
+
+At the end of the chain, the `Sink` kernel reads data from the last pipe and returns it to the host:
+
+``` c++
+void SinkKernel(queue &q, std::array<float, 1> &out_data) {
```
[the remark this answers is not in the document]
jessicadavies-intel: Ok, I will remove the use of std::array here.

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/README.md
```diff
@@ -0,0 +1,151 @@
...
+        [=]() { out_accessor[0] = Pipes::PipeAt<kEngines>::read(); });
+  });
+}
+```
+
+ The following line defines the functionality of each compute unit, and submits each compute unit to the given queue `q`. The number of copies to submit is provided by the first template parameter of `submit_compute_units`, while the second template parameter allows you to give the compute units a shared base name, e.g., `ChainComputeUnit`. The functionality of each compute unit must be specified by a generic lambda expression, that takes a single argument (with type `auto`) that provides an ID for each compute unit:
+
+``` c++
+intelfpga::submit_compute_units<kEngines, ChainComputeUnit>(q, [=](auto ID) {
+    float f = Pipes::PipeAt<ID>::read();
```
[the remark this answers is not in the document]
jessicadavies-intel: I will switch this to auto.

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/README.md
```diff
@@ -0,0 +1,151 @@
...
+  });
+}
+```
+
+At the end of the chain, the `Sink` kernel reads data from the last pipe and returns it to the host:
+
+``` c++
+void SinkKernel(queue &q, std::array<float, 1> &out_data) {
+
+  buffer<float, 1> out_buf(out_data.data(), 1);
```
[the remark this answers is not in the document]
jessicadavies-intel: After removing the use of std::array for a single element, this line becomes:

`buffer<float, 1> out_buf(out_data, 1);`

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/compute_units.cpp
jessicadavies-intel: Fixed.

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/README.md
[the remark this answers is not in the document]
tyoungsc: careful:
```c++
buffer<float, 1> out_buf(&out_data, 1);
```

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/compute_units.cpp
jessicadavies-intel: Both options seemed a bit ugly to me, but I prefer to avoid the use of an unnecessary std::array (even if the output buffer syntax becomes more verbose).

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/compute_units.cpp
jessicadavies-intel: In order to maintain consistency across multiple tutorials, I will postpone making this change.

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/compute_units.cpp
tyoungsc: I agreed

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/compute_units.hpp
```diff
@@ -0,0 +1,37 @@
...
+
+namespace {
+template <class Func, template <std::size_t> class Name, std::size_t index>
+class SubmitOneComputeUnit {
+public:
+  SubmitOneComputeUnit(Func &&f, sycl::queue &q) {
+
+    q.submit([&](sycl::handler &h) {
+      h.single_task<Name<index>>([=] {
+        // verifies that f only takes a single argument
```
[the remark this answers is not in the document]
jessicadavies-intel: Yes, it is the intended comment.

The submit_compute_units function takes a lambda function (representing the kernel that should be duplicated) as an argument. That lambda function needs to satisfy certain properties. One of those properties is that the lambda function should take exactly one argument (the compute unit ID). If not, then the compile will fail at this line.

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/README.md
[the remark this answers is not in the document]
jessicadavies-intel: After my change, out_data is a pointer itself.

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/pipe_array.hpp
```diff
@@ -0,0 +1,43 @@
...
+#else
+  #include <CL/sycl/INTEL/fpga_extensions.hpp>
+#endif
+
+
+#include "pipe_array_internal.hpp"
+
+template <class Id, typename BaseTy, size_t depth, size_t... dims>
+struct PipeArray {
+  PipeArray() = delete;
```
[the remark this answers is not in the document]
jessicadavies-intel: Objects of type PipeArray should never be created. This line makes it impossible to accidentally construct PipeArrays, by disabling the default constructor for PipeArray. Since we didn't add any other constructors, it won't be possible to construct PipeArray at all.

akertesz (approved):
Approving to re-trigger the CI tests

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/pipe_array_internal.hpp
```diff
@@ -0,0 +1,26 @@
...
+// Copyright Intel Corporation
+//
+// SPDX-License-Identifier: MIT
+// =============================================================
+
+namespace {
+template <size_t dim1, size_t... dims>
+struct VerifierDimLayer {
+  template <size_t idx1, size_t... idxs>
+  struct VerifierIdxLayer {
```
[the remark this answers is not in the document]
jessicadavies-intel: Yes, I will add comments here, in the separate change I'm planning to make for pipe array improvements.

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/unroller.hpp
```diff
@@ -0,0 +1,15 @@
...
+// SPDX-License-Identifier: MIT
+// =============================================================
+template <size_t it, size_t end> struct Unroller {
+  template <typename Action> static void Step(const Action &action) {
+    action(std::integral_constant<size_t, it>());
+    Unroller<it + 1, end>::Step(action);
+  }
+};
+
+template <size_t end> struct Unroller<end, end> {
```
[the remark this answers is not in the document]
jessicadavies-intel: I will add more comments to the unroller in the planned separate changelist for pipe_array improvements.

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/unroller.hpp
```diff
@@ -0,0 +1,15 @@
+//==============================================================
+// Copyright Intel Corporation
+//
+// SPDX-License-Identifier: MIT
+// =============================================================
+template <size_t it, size_t end> struct Unroller {
```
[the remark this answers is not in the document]
jessicadavies-intel: I will add more comments to the unroller in the planned separate changelist for pipe_array improvements.

JoeOster (approved):

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/CMakeLists.txt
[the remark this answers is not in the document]
keryell: Amazing... :-)

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/build.ninja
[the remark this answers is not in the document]
keryell: Yes, I understood this by reviewing another example...

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/pipe_array.hpp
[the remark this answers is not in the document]
keryell: Thanks. But what I mean is, explain this in a comment. :-) Keep it for your next improvement. This is a **tutorial**, so anyone interested by FPGA (usually not super C++-savvy) should be able to understand this code.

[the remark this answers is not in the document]
tyoungsc: Is this the definition of a singleton class?

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/compute_units.hpp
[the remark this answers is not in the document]
tyoungsc: I agree with Ronan here. Should the wording be "verifies that f only...". "verifies" is the weird part to me. It just won't compile. These static_asserts may be useful though (https://en.cppreference.com/w/cpp/types/is_invocable). Maybe something like:
```c++
static_assert(std::is_invocable<Func, std::integral_constant<std::size_t, index>>::value, "Your Error Message");
```
This will verify that Func is callable with a single `std::integral_constant<std::size_t, index>` operand.

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/compute_units.cpp
keryell: Otherwise this message will never appear...

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/compute_units.hpp
```diff
@@ -0,0 +1,37 @@
+//==============================================================
+// Copyright Intel Corporation
+//
+// SPDX-License-Identifier: MIT
+// =============================================================
+#include <CL/sycl.hpp>
+#include <utility>
+
+namespace {
+template <class Func, template <std::size_t> class Name, std::size_t index>
```
keryell, line 10: In modern C++, this is
```c++ 
template <typename Func, template <std::size_t> typename Name, std::size_t index>
```

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/compute_units.hpp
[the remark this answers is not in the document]
keryell: Perhaps `// Assume that Func is a callable with a std::size_t argument`?
But I would put this in the comment above describing `SubmitOneComputeUnit`

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/compute_units.hpp
```diff
@@ -0,0 +1,37 @@
...
+class SubmitOneComputeUnit {
+public:
+  SubmitOneComputeUnit(Func &&f, sycl::queue &q) {
+
+    q.submit([&](sycl::handler &h) {
+      h.single_task<Name<index>>([=] {
+        // verifies that f only takes a single argument
+        f(std::integral_constant<std::size_t, index>());
```
keryell, line 18: I am puzzled by this code... What are you trying to do here?
Do you mean
```c++
f(index);
```
or this is an obfuscation contest?

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/pipe_array.hpp
```diff
@@ -0,0 +1,44 @@
...
+  namespace INTEL = sycl::intel;  // Namespace alias for backward compatibility
+#else
+  #include <CL/sycl/INTEL/fpga_extensions.hpp>
+  using namespace sycl;
+#endif
+
+
+#include "pipe_array_internal.hpp"
+
+template <class Id, typename BaseTy, size_t depth, size_t... dims>
```
keryell, line 23: `template <typename Id, typename BaseTy, size_t depth, size_t... dims>`

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/compute_units.hpp
```diff
@@ -0,0 +1,37 @@
...
+template <template <std::size_t> class Name, class Func, std::size_t... index>
+inline constexpr void ComputeUnitUnroller(sycl::queue &q, Func &&f,
+                                          std::index_sequence<index...>) {
+  (SubmitOneComputeUnit<Func, Name, index>(f, q), ...); // fold expression
+}
+} // namespace
+
+// N is the number of compute units
+// Name is the kernel's name
+template <std::size_t N, template <std::size_t ID> class Name, class Func>
```
keryell, line 33: template <std::size_t N, template <std::size_t ID> typename Name, typename Func>

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/compute_units.hpp
```diff
@@ -0,0 +1,37 @@
...
+constexpr void submit_compute_units(sycl::queue &q, Func &&f) {
+  std::make_index_sequence<N> indices;
```
keryell, line 35: Usually I use `boost::hana` to avoid all this complexity, but I guess you do not want to add some external dependency in this project.

keryell (commented):
A few other clarifications...

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/pipe_array.hpp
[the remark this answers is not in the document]
keryell: No. Just that this class is only used as a meta-function for meta-programming and its instantiation has no usage.
The problem here is that since nothing is explained, a user trying to create such an object could not, but not knowing why, since there is no `static_assert` or comment to clarify the reason.
On the other hand, creating such an object has no effect, so at the end this is a lot of work to just add some smoke curtain... Perhaps just removing this line is good? :-)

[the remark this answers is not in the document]
tyoungsc: I agree this is confusing. Having a class called `PipeArray` but you cannot make an object seems, at best, counterintuitive.  I would think I could make as many of these as I want... Maybe we can figure something cleaner out later @jessicadavies-intel

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/compute_units.hpp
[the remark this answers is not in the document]
tyoungsc: Scanning the usage of `submit_compute_units` in submit_compute_units.cpp, this operand is used to index the `PipeArray`. There, it HAS to be a constant, hence the `std::integral_constant`

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/compute_units.hpp
tyoungsc: Operand needs to be a constant. However presumably `index` is already a constant since its a template parameter...

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/compute_units.hpp
tyoungsc: Each project has to be standalone ...  no boost, not even common _shared_ files across projects :(

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/compute_units.hpp
```diff
@@ -0,0 +1,37 @@
...
+inline constexpr void ComputeUnitUnroller(sycl::queue &q, Func &&f,
+                                          std::index_sequence<index...>) {
+  (SubmitOneComputeUnit<Func, Name, index>(f, q), ...); // fold expression
+}
+} // namespace
+
+// N is the number of compute units
+// Name is the kernel's name
+template <std::size_t N, template <std::size_t ID> class Name, class Func>
+constexpr void submit_compute_units(sycl::queue &q, Func &&f) {
```
tyoungsc, line 34: @jessicadavies-intel `submit_compute_units` is a function and therefore I _think_ should be `SubmitComputeUnits`...

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/compute_units.hpp
jessicadavies-intel: Thank you, I will fix this and remember it going forward.

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/pipe_array.hpp
jessicadavies-intel: Will be changed with other pipe_array changes.

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/compute_units.hpp
jessicadavies-intel: Yes, I will change this.

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/compute_units.hpp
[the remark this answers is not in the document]
jessicadavies-intel: Yes, that's right Tanner. The main challenge in writing the compute units API was needing to have a constexpr function argument. The user writes a lambda function containing their kernel's functionality. Within that kernel code, the user should be able to access the compute unit ID as a constexpr.

This was the starting point I found: https://stackoverflow.com/questions/56937863/constexpr-lambda-argument

I will add the static_assert, thanks for the suggestion!

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/compute_units.hpp
jessicadavies-intel: Ok, I tried `f(index)` but it doesn't compile. Good to jog my memory on why all of this was necessary! I will try to give my own explanation here, but this reference might be more helpful: https://stackoverflow.com/questions/56937863/constexpr-lambda-argument

Yes, `index` is constexpr. However, there is actually no way for a function argument to be constexpr. So within the function `f`, even though it was passed a constexpr, in `f` it won't be known to be constexpr.

To get around this limitation, it's necessary to encode the value in a type instead. Then within `f` it's possible to query that type and obtain the value.

`std::integral_constant` is a way to encode a value in a type.

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/compute_units.hpp
jessicadavies-intel: Thank you, I will make this change.

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/compute_units.hpp
keryell: I have found the reason of our disagreement and now I understand the reason for your code. :-)
When I saw old C++ coding style
```c++
         f(std::integral_constant<std::size_t, index>());
```
I actually parsed it:
```c++
         f(std::integral_constant<std::size_t, index>()());
```
so calling `std::integral_constant::operator()` returning the value itself, which is `index` here, according to https://en.cppreference.com/w/cpp/types/integral_constant.

If you had written instead a more modern
```c++
         f(std::integral_constant<std::size_t, index> {});
```
it would have seen immediately it was just creating a `std::integral_constant` and passing it.
The other super tricky part is also later
```c++
       auto f = Pipes::PipeAt<ID>::read();
```
which works because there is an implicit conversion operator `std::integral_constant::operator value_type` returning back a `std::size_t` from `ID`.
Ahem... All this without a single comment in a tutorial? How is a SYCL programmer expected to understand this? :-( This is the most complex code I have seen in oneAPI DPC++ so far... :-)
Another simpler way to do this nowadays could be to use a C++20 templated lambda where you put `index` directly in the template parameter. But actually the code from the end-user point of view would be longer. So your solution seems better!

keryell (commented):
Please express yourself in a lot of comments in the code. :-)

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/compute_units.hpp
jessicadavies-intel: I'm glad we're on the same page! I will apply your tip about using an initializer list to make this more readable.

The tutorials should focus on teaching people how to use SYCL for FPGAs, not how C++ template metaprogramming works. We plan to move these header files into a separate library, so they can be re-used and the implementation details can be ignored. Nevertheless, more comments will also improve the quality of these libraries :-)

keryell: > using an initializer list

No, in this case this is a uniform initialization, not an initializer list, even if both have `{}`. https://isocpp.org/wiki/faq/cpp11-language#uniform-init

Otherwise, I do not think this is ethical to push uncommented open-source code all over the place... I hate write-only code! :-)
If the code is open-source this is to learn from it...

PR #168, opened 2020-09-18
ApplicationDebugger/array-transform: fine-tune README

barisaktemur:
# Description

1. Update the debugger name and add trademark note.

2. Update the compiler name.

3. Fine-tune spacing of the table.

4. Refer to the Get Started Guide for details.

5. Update the examples outputs.

### DIFF
--- a/Tools/ApplicationDebugger/array-transform/README.md
+++ b/Tools/ApplicationDebugger/array-transform/README.md
@@ -1,21 +1,21 @@
 # `array-transform` Sample
 
-This is a small DPC++ example for exercising the application debugger (GDB\*)
-of DPC++.  It is highly recommended that you go through this sample *after*
-you familiarize yourself with the basics of DPC++, and *before* you start
-using the debugger.
+This is a small DPC++ example for exercising application debugging using
+Intel&reg; Distribution for GDB\*.  It is highly recommended that you go
+through this sample *after* you familiarize yourself with the basics of
+DPC++, and *before* you start using the debugger.
 
 This sample accompanies the
 [Get Started Guide](https://software.intel.com/en-us/get-started-with-debugging-dpcpp)
 of the application debugger.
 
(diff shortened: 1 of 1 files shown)


tomlenth:
A couple of comments regarding the README: 
-Before the Example Output section, include a the heading Running the Sample.  this may also be a good place to introduce GDB and how the sample should be used with the debugger. 
- Similarly, the "Useful Commands" section is a great additoin, but needs more context.  I ran into this in the Getting Started Guide for GDB as well.  We don't ever tell the user when/how to start GDB.  It would be ideal to precede this section with something like "Type the following to invoke the debugger," it may seem obvious, but no where in the readme or Getting Started to we tell the user this and instead we just provide commands out of context.  
-lastly, I may try to introduce a more standard section regarding setvars.  although having it step one is front and center, this gets problematic if every sample starts with this since the script will generate warning on subsequent runs.  nothing to change for now, but be aware we may change this.

JoeOster (approved):

JoeOster (changes requested):
Waiting for Toms suggestions to be addressed

anjgola (dismissed):

JoeOster:
@barisaktemur - Please take a look at the comments by TomLenth and address as appropriate

tomlenth:
Baris and I had an offline discussion.  instead of explicitly telling the user to start gdb, we'll point the user to the OS specific GDB getting started guide at the end of each OS specific section of the readme.  This is much more comprehensive and ensures that the user has additional context needed to use the sample.  The link to the GSGs can be positioned as the last step in in build/run section. We've had good luck with this metaphor in the past as the explicit call to action.

JoeOster (approved):

anjgola:
Pending @tomlenth  approval for merge.

tomlenth:
this works for me.  thank you for including this.

PR #174, opened 2020-09-22
Adding FPGA tutorial: shannonization

tyoungsc:

# Description

Adding a new FPGA tutorial: shannonization

## Type of change

- [X] Sample Migration (Moving sample from old repository after completing checklist established)

# How Has This Been Tested?

Tested locally and in our internal regtest infrastructure.
- [X] Command Line
- [X] Visual Studio


akertesz (dismissed):

pmpeter1 (changes requested):
Function naming does not follow style guide (e.g. usage, strStartsWith, submitKernels).

akertesz:
@tomlenth -- I don't seem able to add you as a reviewer, but please take a look.
@AlbyB -- same thing!

tyoungsc:
@pmpeter1 I changed those function names.

tomlenth:
Readme looks great.  thank you.

tomlenth:
Please check the link to the binary.  currently it is giving me a 404.

tyoungsc:
> Please check the link to the binary. currently it is giving me a 404.

I think this is expected, right @akertesz ?

pmpeter1 (changes requested):
You missed countSorted

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/src/shannonization.cpp
```diff
@@ -0,0 +1,335 @@
...
+    if (golden_n != n) {
+      success = false;
+      std::cerr << "ERROR: Kernel version " << Version << " output is incorrect"
+                << " (Expected=" << golden_n << ", Result=" << n << ")\n";
+    }
+
+    // get profiling info
+    auto start = e.get_profiling_info<info::event_profiling::command_start>();
+    auto end = e.get_profiling_info<info::event_profiling::command_end>();
+    kernel_latency[i] = (end - start) / 1000000000.0f;
```
keryell, line 170:
```suggestion
    kernel_latency[i] = (end - start) / 1e9;
```
I R=removed also the `f` since the result is a... `double`!

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/src/shannonization.cpp
```diff
@@ -0,0 +1,335 @@
...
+  }
+
+  // If all the iterations were successful, print the throughput results.
+  // The FPGA emulator does not accurately represent the hardware performance
+  // so we don't print performance results when running with the emulator
+  if(success) {
```
keryell, line 176:
```suggestion
  if (success) {
```

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/src/shannonization.cpp
```diff
@@ -0,0 +1,335 @@
...
+      } else if (StrStartsWith(arg, "--B=")) {
+        b_size = std::stoi(str_after_equals);
+      } else {
+        std::cout << "WARNING: ignoring unknown argument '" << arg << "'\n";
+      }
+    }
+  }
+
+  // ensure the arrays have more than 3 elements
+  if(a_size <= 3) {
```
keryell, line 229:
```suggestion
  if (a_size <= 3) {
```

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/src/shannonization.cpp
```diff
@@ -0,0 +1,335 @@
...
+    std::cout << "WARNING: array A must have more than 3 "
+                  "elements, increasing its size\n";
```
keryell, line 231: For `std::cout` use `std::endl` instead of `\n` everywhere. Otherwise you might never see the output if the program crashes...

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/src/shannonization.cpp
```diff
@@ -0,0 +1,335 @@
...
+    a_size = 4;
+  }
+  if(b_size <= 3) {
```
keryell, line 234:
```suggestion
  if (b_size <= 3) {
```

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/src/shannonization.cpp
```diff
@@ -0,0 +1,335 @@
...
+  }
+
+  std::cout << "Generating input data\n";
+
+  // seed the random number generator
+  srand(777);
+
+  // initialize input data
+  std::vector<unsigned int> a(a_size), b(b_size);
+  std::generate(a.begin(), a.end(), [n = 0]() mutable { return n++; });
```
keryell, line 253: `std::iota`

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/src/shannonization.cpp
```diff
@@ -0,0 +1,335 @@
...
+  std::generate(b.begin(), b.end(), [=]() mutable { return rand() % a_size; });
```
keryell, line 254: Why `mutable`?

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/README.md
```diff
@@ -0,0 +1,318 @@
...
+![](basic.png)
+
+The goal of the shannonization optimization is to remove operations from the critical path. In this case, we can precompute the _next_ value of `c` (fittingly named `c_next`) for a latter iteration of the loop to use when needed (i.e. the next time `A[i] < v`). This optimization is shown in the code below.
+
+```c++
+int A[SIZE] = {/*...*/};
+int v = /*some dynamic value*/
+int c = 0;
+int c_next = 1;
+for(int i = 0; i < SIZE; i++) {
```
keryell, line 40:
```suggestion
for (int i = 0; i < SIZE; i++) {
```

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/README.md
```diff
@@ -0,0 +1,318 @@
...
+  if(A[i] < v) {
```
keryell, line 41:
```suggestion
  if (A[i] < v) {
```

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/README.md
```diff
@@ -0,0 +1,318 @@
...
+  int A_count, B_count;
+  intersection_size = 0;
+
+  // initialize first elements
+  a = APipe::read();
+  b = BPipe::read();
+  A_count = 1;
+  B_count = 1;
+
+  while(A_count < A_size || B_count < B_size) {
```
keryell, line 89:
```suggestion
  while (A_count < A_size || B_count < B_size) {
```

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/README.md
```diff
@@ -0,0 +1,318 @@
...
+    // values match increment counter
+    if(a == b) {
```
keryell, line 91:
```suggestion
    if (a == b) {
```

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/README.md
```diff
@@ -0,0 +1,318 @@
...
+      intersection_size++;
+    }
+
+    // read in new element
+    if(a < b && A_count < A_size) {
```
keryell, line 96:
```suggestion
    if (a < b && A_count < A_size) {
```

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/README.md
```diff
@@ -0,0 +1,318 @@
...
+      a = APipe::read();
+      A_count++;
+    } else if(B_count < B_size) {
```
keryell, line 99:
```suggestion
    } else if (B_count < B_size) {
```

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/README.md
```diff
@@ -0,0 +1,318 @@
...
+      b = BPipe::read();
+      B_count++;
+    }
+  };
+
+  // check the last values
+  if(a == b) {
```
keryell, line 106:
```suggestion
  if (a == b) {
```

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/README.md
```diff
@@ -0,0 +1,318 @@
...
+     make report
+     ```
+   * Compile for FPGA hardware (longer compile time, targets FPGA device): 
+     ```
+     make fpga
+     ```
+3. (Optional) As the above hardware compile may take several hours to complete, an Intel® PAC with Intel Arria® 10 GX FPGA precompiled binary can be downloaded <a href="https://software.intel.com/content/dam/develop/external/us/en/documents/shannonization.fpga.tar.gz" download>here</a>.
+ 
+### On a Windows* System
+Note: `cmake` is not yet supported on Windows. A build.ninja file is provided instead.
```
keryell, line 160: Ah, I see why there is some Ninja code.
Curious this CMake configuration is not working yet...

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/README.md
```diff
@@ -0,0 +1,318 @@
...
+
+#### Version 0
+The first version of the kernel, `Intersection<0>`, is the baseline implementation of the intersection kernel. Check the *Details* pane in the *Loop Analysis* tab for the `while` loop in the `Intersection<0>` kernel. You will notice that the *Block Scheduled f<sub>MAX</sub>* for the `Intersection<0>` kernel is far lower than the target (e.g. ~140 MHz). The *Details* pane shows that the most critical path contains the operations mentioned earlier at the end of the [Algorithm Details](#algorithm-details) Section.
+
+#### Version 1
+The second version of the kernel, `Intersection<1>`, uses the shannonization optimization to remove the increment of `A_count` and `B_count` from the critical path. To do this, we create two new variables, `A_count_next` and `B_count_next` which will store the value of `A_count` and `B_count` **for the next iteration of the loop**. The code snippet below shows how `A_count` and `B_count` are updated using `A_count_next` and `B_count_next`:
+```c++
+  ...
+  if(a < b && A_count < A_size) {
```
keryell, line 201:
```suggestion
  if (a < b && A_count < A_size) {
```

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/README.md
```diff
@@ -0,0 +1,318 @@
...
+    a = APipe::read();
+    A_count = A_count_next;
+    A_count_next++;
+  } else if(B_count < B_size) {
```
keryell, line 205:
```suggestion
  } else if (B_count < B_size) {
```

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/README.md
```diff
@@ -0,0 +1,318 @@
...
+```
+
+Now, the assignments of `A_count = A_count_next` and `B_count = B_count_next` can be done **in parallel to the increment of the counts for the next iteration of the loop** (i.e. in parallel to `A_count_next++` and `B_count_next++`). This removes the 32-bit Integer Add Operation from the critical path, as can be seen in the *Details* pane of the *Loop Analysis* report for the `Intersection<1>` kernel. The *Loop Analysis* pane will show an increase in the *Block Scheduled f<sub>MAX</sub>* (e.g. ~190 MHz).
+
+#### Version 2
+The third version of the kernel, `Intersection<2>`, extends the previous optimization by removing the 32-bit Integer Compare Operation from the critical path. The first step is to precompute the comparisons for the next loop iteration (`A_count_inrange` and `B_count_inrange`) and the next iteration (`A_count_next_inrange` and `B_count_next_inrange`), as we did for the accumulations. This is shown in the code snippet below:
+
+```c++
+  ...
+  if(a < b && A_count_inrange) {
```
keryell, line 220:
```suggestion
  if (a < b && A_count_inrange) {
```

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/README.md
```diff
@@ -0,0 +1,318 @@
...
+    a = APipe::read();
+
+    A_count = A_count_next;
+    A_count_next++;
+
+    A_count_inrange = A_count_next_inrange;
+    A_count_next_inrange = A_count_next < A_size;
+  } else if(B_count_inrange) {
```
keryell, line 228:
```suggestion
  } else if (B_count_inrange) {
```

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/README.md
```diff
@@ -0,0 +1,318 @@
...
+    B_count_inrange = B_count_next_inrange;
+    B_count_next_inrange = B_count_next < B_size;
+  }
+  ...
+```
+However, this places a 32-bit Integer Add Operation back into the critical path (e.g. `A_count_next++` must computed before before computing `A_count_next_inrange = A_count_next < A_size`). To remove this addition from the critical path, we do the same optimization as version 1. That is, we now precompute the additions for the next **two** iterations of the loop (`A_count_next_next` and `B_count_next_next`), which again removes the addition from the critical path. This is shown in the code snippet below:
+
+```c++
+  ...
+  if(a < b && A_count_inrange) {
```
keryell, line 243:
```suggestion
  if (a < b && A_count_inrange) {
```

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/README.md
```diff
@@ -0,0 +1,318 @@
...
+    a = APipe::read();
+
+    A_count_inrange = A_count_next_inrange;
+    A_count_next_inrange = A_count_next_next < A_size;
+
+    A_count = A_count_next;
+    A_count_next = A_count_next_next;
+    A_count_next_next++;
+  } else if(B_count_inrange) {
```
keryell, line 252:
```suggestion
  } else if (B_count_inrange) {
```

tyoungsc:
> 2.8 from 2009 seems unlikely...

This is file is fairly consistent across all FPGA samples. Is there a suggested value for this??

keryell (changes requested):
Interesting FPGA use case.
Just a few nitpicks.

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+set(CMAKE_CXX_COMPILER "dpcpp")
+
+cmake_minimum_required (VERSION 2.8)
```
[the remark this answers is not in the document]
tyoungsc: What is the suggested minimum version? Should we be consistent? As far as I can see, all of our FPGA tutorials set this as the minimum...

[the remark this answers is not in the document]
akertesz: All of the FPGA samples use this as the minimum version. I believe it is for compatibility with RHEL7, which which not an officially supported OS is required by some of our flows.

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/src/shannonization.cpp
```diff
@@ -0,0 +1,335 @@
...
+
+//
+// helper to check if string 'str' starts with 'prefix'
+//
+bool StrStartsWith(std::string& str, std::string prefix) {
+  return str.find(prefix) == 0;
+}
+
+//
+// helper to count instances of an element 'x' in a sorted vector 'v'
```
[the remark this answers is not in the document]
tyoungsc: To be honest, I added it because I got tired of waiting for hardware runs to generate golden results. I will add a comment.

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/src/shannonization.cpp
akertesz: I don't think this is true. If the program catches, the exception handler will call std::terminate() which should flush the buffer. Also, the use of \n rather than std::endl is mandated by the coding standard for code samples. I believe that if an explicit flush is desired (of which I'm not convinced here) that an explicit std::flush is preferred.

tyoungsc: Our internal guidelines say to do the exact opposite. @akertesz ??

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/src/shannonization.cpp
```diff
@@ -0,0 +1,335 @@
...
+  std::cout
+      << "USAGE: ./intersection [--A=<size of list A>] [--B=<size of list B>]"
+      << "[--iters=<number of times to run the kernel>] [-h --help]\n";
+}
+
+//
+// helper to check if string 'str' starts with 'prefix'
+//
+bool StrStartsWith(std::string& str, std::string prefix) {
+  return str.find(prefix) == 0;
```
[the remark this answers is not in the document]
tyoungsc: I used it in a bunch of my designs to parse input args. Ideally, we would have some centralized input args parsing mechanism like [boost](https://www.boost.org/doc/libs/1_74_0/doc/html/program_options.html)

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/src/shannonization.cpp
```diff
@@ -0,0 +1,335 @@
...
+  // input sizes
+  const int a_size = a.size();
+  const int b_size = b.size();
+
+  // setup the input buffers
+  buffer a_buf(a);
+  buffer b_buf(b);
+
+  // setup the output buffer
+  buffer<int, 1> n_buf(&n, 1);
```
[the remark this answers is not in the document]
tyoungsc: breaks my compile

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/src/shannonization.cpp
```diff
@@ -0,0 +1,335 @@
...
+
+  // submit the kernel that produces table A
+  q.submit([&](handler& h) {
+    accessor a_accessor { a_buf, h, read_only };
+    h.single_task<ProducerA<Version>>([=]() [[intel::kernel_args_restrict]] {
```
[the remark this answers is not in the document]
tyoungsc: breaks my compile

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/src/shannonization.cpp
```diff
@@ -0,0 +1,335 @@
...
+      for (int i = 0; i < a_size; i++) {
+        ProduceAPipe::write(a_accessor[i]);
+      }
```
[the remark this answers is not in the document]
tyoungsc: tried it, breaks my compile

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/src/shannonization.cpp
```diff
@@ -0,0 +1,335 @@
...
+    });
+  });
+
+  // submit the kernel that produces table B
+  q.submit([&](handler& h) {
+    accessor b_accessor { b_buf, h, read_only };
+    h.single_task<ProducerB<Version>>([=]() [[intel::kernel_args_restrict]] {
+      for (int i = 0; i < b_size; i++) {
+        ProduceBPipe::write(b_accessor[i]);
+      }
```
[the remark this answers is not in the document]
tyoungsc: breaks my compile

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/src/shannonization.cpp
```diff
@@ -0,0 +1,335 @@
...
+    });
+  });
+
+  // submit the kernel that performs the intersection
+  event e = q.submit([&](handler& h) {
+    // output accessor
+    accessor n_accessor { n_buf, h, write_only, noinit };
+
+    h.single_task<Worker<Version>>([=]() [[intel::kernel_args_restrict]] {
```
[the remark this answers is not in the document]
tyoungsc: breaks my compile

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/src/shannonization.cpp
```diff
@@ -0,0 +1,335 @@
+//==============================================================
+//Copyright Intel Corporation
+//
+// SPDX-License-Identifier: MIT
+// =============================================================
+#include <CL/sycl.hpp>
```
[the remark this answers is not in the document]
tyoungsc: breaks my compile, cannot find file. All of our FPGA designs use CL/sycl.hpp. @akertesz ??

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/src/IntersectionKernel.hpp
```diff
@@ -0,0 +1,321 @@
+#ifndef __INTERSECTIONKERNEL_HPP__
+#define __INTERSECTIONKERNEL_HPP__
+
+#include <CL/sycl.hpp>
```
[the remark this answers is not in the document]
tyoungsc: Same. Breaks compile.

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/src/shannonization.cpp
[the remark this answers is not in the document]
akertesz: The location of the header in our oneAPI install is in the CL directory. On my reference beta09 install, /opt/intel/oneapi/compiler/2021.1-beta09/linux/include/sycl/CL/sycl.hpp.

I am not aware of any plans to move the header location in upcoming releases.

akertesz:
@tyoungsc , @tomlenth  -- The broken link is expected, unfortunately. The precompiled binaries are only uploaded once per beta, so the link will unbreak later on. I'm working on a better way to handle this. It's not a dealbreaker because this is just a convenience. You can always wait an hour or two and compile your own binary.

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/src/shannonization.cpp
keryell: > I don't think this is true. If the program catches, the exception handler will call std::terminate() which should flush the buffer.

Can you show me where this is stated in the C++ standard?

> Also, the use of \n rather than std::endl is mandated by the coding standard for code samples. 

Strange... @erichkeane I am curious about who writes them...

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/src/shannonization.cpp
```diff
@@ -0,0 +1,335 @@
...
+#include <algorithm>
+#include <vector>
+#include <type_traits>
```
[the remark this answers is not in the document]
keryell: My point is also to put `vector` after `type_traits`

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/src/shannonization.cpp
```diff
@@ -250,8 +253,8 @@ int main(int argc, char** argv) {
 
   // initialize input data
   std::vector<unsigned int> a(a_size), b(b_size);
-  std::generate(a.begin(), a.end(), [n = 0]() mutable { return n++; });
-  std::generate(b.begin(), b.end(), [=]() mutable { return rand() % a_size; });
+  std::iota(a.begin(), a.end(), 0);
+  std::generate(b.begin(), b.end(), [=]() { return rand() % a_size; });
```
keryell, line 257:
```suggestion
  std::generate(b.begin(), b.end(), [=] { return rand() % a_size; });
```
Be lazy :-)

keryell (changes requested):
Thanks for the cleaning.
Just a few details...

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/src/IntersectionKernel.hpp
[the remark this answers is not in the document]
keryell: Ah interesting... @bader A bug in DPC++? I thought it was SYCL 2020 by default now.

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/src/shannonization.cpp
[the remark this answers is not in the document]
keryell: Ah, because of the attribute... :-(

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/src/shannonization.cpp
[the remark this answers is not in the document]
keryell: Curious... @bader Did we forget some deduction guide in the spec?

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/src/shannonization.cpp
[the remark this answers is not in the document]
keryell: Ah, interesting feedback! So complexity analysis is really useful sometime... :-)

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/src/shannonization.cpp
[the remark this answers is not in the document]
keryell: @jbrodman I thought we added a lot of goodies in SYCL 2020, like a container-like behaviour to the accessor...

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/src/shannonization.cpp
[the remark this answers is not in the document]
keryell: SYCL 2020 is no longer based on OpenCL. This is why the headers are now moving to `SYCL` according to the specification. `CL/sycl.hpp` still works as a compatibility layer. @bader

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/src/IntersectionKernel.hpp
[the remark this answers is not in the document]
bader: DPC++ sets SYCL-2020 by default, but the implementation is not fully conformant yet. Unfortunately new include path is not supposed yet.

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/src/shannonization.cpp
erichkeane: In cout, \n causes a flush as far as I know, which is why the coding  standard recommends doing  '\n'. As far as std::terminate flushing the buffer, I'm not aware of that being a standards requirement, but I can look closer tomorrow.

keryell: No emergency. :-) I think `\n` causes a flush in C, not in C++... Perhaps this is the source of the confusion in the coding guideline?

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/src/shannonization.cpp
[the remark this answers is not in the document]
bader: @tyoungsc, could you try to build with C++17 mode, please?

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/src/shannonization.cpp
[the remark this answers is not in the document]
tyoungsc: Not just in algos class!

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/src/shannonization.cpp
erichkeane: The '\n' flushing for cout is an implementation behavior that all major ones share.  

See here for ifo about the terminate flushing: https://en.cppreference.com/w/cpp/io/manip/flush

"In most other usual interactive I/O scenarios, std::endl is redundant when used with std::cout because any input from std::cin, output to std::cerr, or program termination forces a call to std::cout.flush(). "

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/src/shannonization.cpp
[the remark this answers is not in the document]
tyoungsc: Are default is C++17. I tried adding the flag specifically, still didn't work.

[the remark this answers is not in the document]
tyoungsc: Between
`buffer<int, 1> n_buf(&n, 1);`
and
`buffer n_buf(&n, range<1>(1));`

My preference would be the former -- I can instantly tell what type my buffer is and it is (marginally) less verbose.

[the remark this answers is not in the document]
keryell: > @keryell, probably we need more rules to make your version work.

So it looks like this is a SYCL 2020 spec bug...
I created https://gitlab.khronos.org/sycl/Specification/-/issues/453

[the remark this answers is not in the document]
tyoungsc: Sounds good. For now, I will leave this as-is.

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/src/shannonization.cpp
keryell: Yes but none of these applies. In the case of a crash or `std::terminate` (aka `std::abort()` on Unix by default!), there is no flush.
Since nobody believes me or https://isocpp.org/wiki/faq/input-output#endl-vs-slash-n, here is the proof:
https://godbolt.org/z/68GWsE

keryell: The problem with coding guide-lines for **tutorials** is that there should not inoculate more bad coding practices in customer code... Life is already hard enough like that... :-/

tyoungsc: Based on our internal guidelines, which state:
> For line endings, use '\n' over std::endl (unless you actually need the explicit flush as well as the newline).

I believe the compromise here would be to continue using '\n' BUT place a `std::flush()` before the call to `std::terminate`.

What do you think @akertesz and @keryell??

akertesz: Yes, that was my proposal. 

Any exception in the tutorial code will be handled by either the asynchronous handler, or caught by the synchronous try-catch block. I just checked the asynchronous handler in dpc_common.hpp, and saw that it uses std::endl. If we add an explicit flush in the synchronous catch block before the std::terminate, we're covered.

Personally I prefer std::flush to std::endl as it is more explicit, but probably the correct thing is to follow the style in dpc_common.hpp.

akertesz (approved):
Thanks Tanner, I'll relaunch CI now.

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/shannonization/src/shannonization.cpp
keryell: I believe that I am right now in a TV reality show about code obfuscation.
Are you SERIOUS??? `\n` + `std::flush`?
Why not copy-pasting the standard C++ library to be sure we are explicit... ;-)
And add a few more `std::flush` just in case... :-)

tyoungsc: I am no code standard guru, but having `std::endl` all over the place seems a little verbose, therefore, I am a fan of `\n`. That being said, I think in the case of the catch block, perhaps we should be using `std::cerr` in the catch block? That way the messages are flushed?

keryell: Yes, `std::cerr` is the way to go here anyway.

PR #213, opened 2020-10-02
oneMKL sample updates for beta10

petercad:
# Description

Updates existing oneMKL samples for compatibility with beta10:
- [X] updated header locations
- [X] LAPACK exception changes

Changes retain beta09 compatibility. At the same time, old beta08 compatibility code has been dropped.

## Type of change

# How Has This Been Tested?

- [X] Command Line

### DIFF
--- a/Libraries/oneMKL/block_cholesky_decomposition/solve.cpp
+++ b/Libraries/oneMKL/block_cholesky_decomposition/solve.cpp
@@ -37,11 +37,14 @@
 #include <iostream>
 #include <vector>
 #include "mkl.h"
-#include "mkl_sycl.hpp"
 
-// Temporary code for beta08 compatibility. oneMKL routines
-//  move to the oneapi namespace in beta09.
-namespace oneapi {}
+#if __has_include("oneapi/mkl.hpp")
+#include "oneapi/mkl.hpp"
+#else
+// Beta09 compatibility -- not needed for new code.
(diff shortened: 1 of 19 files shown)


Libraries/oneMKL/block_cholesky_decomposition/solve.cpp
```diff
@@ -69,7 +72,7 @@ int main() {
             } catch(mkl::lapack::exception const& e) {
                 // Handle LAPACK related exceptions happened during asynchronous call
                 info = e.info();
-                std::cout << "Unexpected exception caught during asynchronous LAPACK operation:\n" << e.reason() << "\ninfo: " << e.info() << std::endl;
```
JoeOster, line 72: You can use namespace std, thus eliminating most if not all std::cout
std:endl should still be used in exceptions

JoeOster (changes requested):

Libraries/oneMKL/block_cholesky_decomposition/solve.cpp
petercad: Thanks for the review! Since the oneMKL samples still need to undergo a comprehensive style update -- it is OK to make this change with the other style changes as part of a separate PR?

JoeOster: I can live with it, but it must be done prior to going gold

PR #226, opened 2020-10-06
Insert image for wiki

JoeOster:
# Description

Adding a "image" folder and an image for the wiki article on Folder structure.


tomlenth (approved):

.github/images/README.md
```diff
@@ -0,0 +1 @@
+
```
anjgola, line 1: I think this already exists. As you can see in the existing wiki there are several images in use.

The wiki content is not stored in the repository branches to avoid cluttering that branch when cloning the repository itself. Github wikis are like a separate webpage. They have their own 'virtual' repo that can be cloned by doing : 

$ git clone https://github.com/YOUR_USERNAME/YOUR_REPOSITORY.wiki.git

In that there is already a images folder IIRC.

anjgola: At the bottom of the 'wiki' tab the URL is listed : https://github.com/oneapi-src/oneAPI-samples.wiki.git

PR #239, opened 2020-10-13
Update header location and namespace for beta10

akertesz:
# Description

In beta10, the Intel header locations and the Intel SYCL extension namespace changed from "intel" to "INTEL" following the Khronos SYCL requirements. There is no deprecation period. This minor change updates the vector-add sample for beta10 compatibility.

This will fail on beta09. I consider this acceptable for master, since beta10 release tag is only 2 weeks away.

While in the neighborhood, I also fixed up the formatting of the async handler code.

## Type of change

# How Has This Been Tested?

- [X] Command Line


JoeOster (approved):
you can use std namespace as well as sycl, but in this case it won't change a lot

akertesz:
I'm not the sample owner, so I don't want to touch anything material about the sample. I'm just fixing the beta10 failure for Jennifer. And I figured I'd tidy up the formatting issues while I was here.

akertesz:
All of the validation failures are as expected

PR #246, opened 2020-10-14
Adding FPGA tutorial 'zero_copy_data_transfer' (after beta10)

tyoungsc:

# Description
Adding new FPGA tutorial on how to use restricted USM to do zero-copy data transfer between the CPU and FPGA.

## Type of change

# How Has This Been Tested?
- [X] Command Line
- [X] Visual Studio

# Testing pipeline
Expected to fail:
1) INTEL/intel namespace/filename incompatibility of beta09/beta10
2) Linux OS support - Ubuntu 20.4 not yet supported for FPGA

### DIFF
--- a/DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/zero_copy_data_transfer/src/buffer_kernel.hpp
+++ b/DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/zero_copy_data_transfer/src/buffer_kernel.hpp
@@ -1,0 +1,55 @@
+#ifndef __BUFFER_KERNEL_HPP__
+#define __BUFFER_KERNEL_HPP__
+#pragma once
+
+#include <vector>
+
+#include <CL/sycl.hpp>
+#include <CL/sycl/INTEL/fpga_extensions.hpp>
+
+using namespace sycl;
+using namespace std::chrono;
+
+// Declare the kernel class name globally to avoid name mangling.
+class BufferWorker;
+
+// kernel using buffers to transfer data
+template <typename T>
+double BufferKernel(queue& q, std::vector<T>& in, std::vector<T>& out,
+                    const unsigned int size) {
+  // start timer
+  auto start = high_resolution_clock::now();
+
+  {
+    // set up the input/output buffers
+    buffer in_buf(in);
+    buffer out_buf(out);
+
+    // launch the computation kernel
+    auto kernel_event = q.submit([&](handler& h) {
+      accessor in_accessor { in_buf, h, read_only };
+      accessor out_accessor { out_buf, h, write_only, noinit };
+
+      h.single_task<BufferWorker>([=]() [[intel::kernel_args_restrict]] {
+        for (size_t i = 0; i < size; i++) {
+          out_accessor[i] = in_accessor[i] * i;
+        }
+      });
+    });
+  }
+
+  // We use the scope above to synchronize the FPGA kernels.
+  // Exiting the scope will cause the buffer destructors to be called
+  // which will wait until the kernels finish and copy the data back to the
+  // host (if the buffer was written to).
+  // Therefore, at this point in the code, we know the kernels have finished
+  // and the data has been transferred back to the host.
+
+  // stop the timer
+  auto end = high_resolution_clock::now();
+  duration<double, std::milli> diff = end - start;
+
+  return diff.count();
+}
+
+#endif /* __BUFFER_KERNEL_HPP__ */
--- a/DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/zero_copy_data_transfer/src/buffer_kernel.hpp
+++ b/DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/zero_copy_data_transfer/src/buffer_kernel.hpp
@@ -27,12 +27,12 @@ double BufferKernel(queue& q, std::vector<T>& in, std::vector<T>& out,
 
     // launch the computation kernel
     auto kernel_event = q.submit([&](handler& h) {
-      accessor in_accessor { in_buf, h, read_only };
-      accessor out_accessor { out_buf, h, write_only, noinit };
+      accessor in_a(in_buf, h, read_only);
+      accessor out_a(out_buf, h, write_only, noinit);
 
       h.single_task<BufferWorker>([=]() [[intel::kernel_args_restrict]] {
         for (size_t i = 0; i < size; i++) {
-          out_accessor[i] = in_accessor[i] * i;
+          out_a[i] = in_a[i] * i;
         }
       });
     });
--- a/DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/zero_copy_data_transfer/src/restricted_usm_kernel.hpp
+++ b/DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/zero_copy_data_transfer/src/restricted_usm_kernel.hpp
@@ -67,7 +67,7 @@ event SubmitProducer(queue& q, T* in_data, size_t size) {
       host_ptr<T> h_in_data(in_data);
 
       for (size_t i = 0; i < size; i++) {
-        const T data_from_host_memory = *(h_in_data + i);
+        T data_from_host_memory = *(h_in_data + i);
         ProducePipe<T>::write(data_from_host_memory);
       }
     });
@@ -87,8 +87,8 @@ event SubmitWorker(queue& q, size_t size) {
   auto e = q.submit([&](handler& h) {
     h.single_task<RestrictedUSM>([=]() [[intel::kernel_args_restrict]] {
       for (size_t i = 0; i < size; i++) {
-        const T data = ProducePipe<T>::read();
-        const T value = data * i; // perform computation
+        T data = ProducePipe<T>::read();
+        T value = data * i; // perform computation
         ConsumePipe<T>::write(value);
       }
     });
@@ -110,7 +110,7 @@ event SubmitConsumer(queue& q, T* out_data, size_t size) {
 
     h.single_task<Consumer>([=]() [[intel::kernel_args_restrict]] {
       for (size_t i = 0; i < size; i++) {
-        const T data_to_host_memory = ConsumePipe<T>::read();
+        T data_to_host_memory = ConsumePipe<T>::read();
         *(h_out_data + i) = data_to_host_memory;
       }
     });
--- a/DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/zero_copy_data_transfer/src/restricted_usm_kernel.hpp
+++ b/DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/zero_copy_data_transfer/src/restricted_usm_kernel.hpp
@@ -1,0 +1,142 @@
+#ifndef __RESTRICTED_USM_KERNEL_HPP__
+#define __RESTRICTED_USM_KERNEL_HPP__
+#pragma once
+
+#include <vector>
+
+#include <CL/sycl.hpp>
+#include <CL/sycl/INTEL/fpga_extensions.hpp>
+
+using namespace sycl;
+using namespace std::chrono;
+
+//
+// The structure of the kernels in this design is shown in the diagram below.
+// The Producer kernel reads the data from CPU memory (via PCIe), producing it
+// for the RestrictedUSM via a pipe. The Worker does the computation on the
+// input data and writes it to the ConsumePipe. The consumer reads the data
+// from this pipe and writes the output back to the CPU memory (via PCIe).
+//
+//                                |----------------------------------|
+//                                |              FPGA                |
+//               |-------------|  |                                  |
+//               |             |  | |- --------|   |---------------| |
+//  |-------|    |             |--->| Producer |==>|               | |
+//  |       |    |             |  | |----------|   |               | |
+//  |  CPU  |<-->| Host Memory |  |                | RestrictedUSM | |
+//  |       |    |             |  | |----------|   |               | |
+//  |-------|    |             |<---| Consumer |<==|               | |
+//               |             |  | |----------|   |---------------| |
+//               |-------------|  |                                  |
+//                                |----------------------------------|
+//
+//
+// As shown in the image above and the code below, we have split this design
+// into three kernels:
+//    1) Producer
+//    2) RestrictedUSM
+//    3) Consumer
+// We do this to decouple the reads/writes from/to the Host Memory over PCIe.
+// Decoupling the memory accesses and using SYCL pipes with a substantial
+// depth ('kPipeDepth' below) allows the kernel to be more resilient against
+// stalls while reading/writing from/to the Host Memory over PCIe. 
+//
+
+// Declare the kernel class names globally to avoid name mangling.
+class RestrictedUSM;
+class Producer;
+class Consumer;
+
+// pipes
+constexpr size_t kPipeDepth = 64;
+template <typename T>
+using ProducePipe = pipe<class ProducePipeClass, T, kPipeDepth>;
+template <typename T>
+using ConsumePipe = pipe<class ConsumePipeClass, T, kPipeDepth>;
+
+//
+// reads the input data from the hosts memory
+// and writes it to the ProducePipe
+//
+template <typename T>
+event SubmitProducer(queue& q, T* in_data, size_t size) {
+  auto e = q.submit([&](handler& h) {
+    h.single_task<Producer>([=]() [[intel::kernel_args_restrict]] {
+      // using a host_ptr tells the compiler that this pointer lives in the
+      // hosts address space
+      host_ptr<T> h_in_data(in_data);
+
+      for (size_t i = 0; i < size; i++) {
+        const T data_from_host_memory = *(h_in_data + i);
+        ProducePipe<T>::write(data_from_host_memory);
+      }
+    });
+  });
+
+  return e;
+}
+
+//
+// The worker kernel in the device:
+//  1) read input data from the ProducePipe
+//  2) perform computation
+//  3) write the output data to the ConsumePipe
+//
+template <typename T>
+event SubmitWorker(queue& q, size_t size) {
+  auto e = q.submit([&](handler& h) {
+    h.single_task<RestrictedUSM>([=]() [[intel::kernel_args_restrict]] {
+      for (size_t i = 0; i < size; i++) {
+        const T data = ProducePipe<T>::read();
+        const T value = data * i; // perform computation
+        ConsumePipe<T>::write(value);
+      }
+    });
+  });
+
+  return e;
+}
+
+//
+// reads output data from the device via ConsumePipe
+// and writes it to the hosts memory
+//
+template <typename T>
+event SubmitConsumer(queue& q, T* out_data, size_t size) {
+  auto e = q.submit([&](handler& h) {
+    // using a host_ptr tells the compiler that this pointer lives in the
+    // hosts address space
+    host_ptr<T> h_out_data(out_data);
+
+    h.single_task<Consumer>([=]() [[intel::kernel_args_restrict]] {
+      for (size_t i = 0; i < size; i++) {
+        const T data_to_host_memory = ConsumePipe<T>::read();
+        *(h_out_data + i) = data_to_host_memory;
+      }
+    });
+  });
+
+  return e;
+}
+
+template <typename T>
+double RestrictedUSMKernel(queue& q, T* in, T* out, size_t size) {
+  // start the timer
+  auto start = high_resolution_clock::now();
+
+  // start the kernels
+  auto worker_event = SubmitWorker<T>(q, size);
+  auto producer_event = SubmitProducer<T>(q, in, size);
+  auto consumer_event = SubmitConsumer<T>(q, out, size);
+
+  // wait for all the kernels to finish
+  q.wait();
+
+  // stop the timer
+  auto end = high_resolution_clock::now();
+  duration<double, std::milli> diff = end - start;
+
+  return diff.count();
+}
+
+#endif /* __RESTRICTED_USM_KERNEL_HPP__ */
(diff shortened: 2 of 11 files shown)


DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/zero_copy_data_transfer/src/buffer_kernel.hpp
```diff
@@ -0,0 +1,55 @@
...
+  // We use the scope above to synchronize the FPGA kernels.
+  // Exiting the scope will cause the buffer destructors to be called
+  // which will wait until the kernels finish and copy the data back to the
+  // host (if the buffer was written to).
+  // Therefore, at this point in the code, we know the kernels have finished
+  // and the data has been transferred back to the host.
+
+  // stop the timer
+  auto end = high_resolution_clock::now();
+  duration<double, std::milli> diff = end - start;
```
pmpeter1, line 50: We should propose an extension of dpc_common.hpp timers such that you can ask for timer.Elapsed(std::milli);

This would allow use of the common headers.

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/zero_copy_data_transfer/src/restricted_usm_kernel.hpp
```diff
@@ -0,0 +1,142 @@
...
+template <typename T>
+event SubmitProducer(queue& q, T* in_data, size_t size) {
+  auto e = q.submit([&](handler& h) {
+    h.single_task<Producer>([=]() [[intel::kernel_args_restrict]] {
+      // using a host_ptr tells the compiler that this pointer lives in the
+      // hosts address space
+      host_ptr<T> h_in_data(in_data);
+
+      for (size_t i = 0; i < size; i++) {
+        const T data_from_host_memory = *(h_in_data + i);
```
pmpeter1, line 70: Why not h_in_data[i]  ?

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/zero_copy_data_transfer/src/restricted_usm_kernel.hpp
```diff
@@ -0,0 +1,142 @@
...
+event SubmitConsumer(queue& q, T* out_data, size_t size) {
+  auto e = q.submit([&](handler& h) {
+    // using a host_ptr tells the compiler that this pointer lives in the
+    // hosts address space
+    host_ptr<T> h_out_data(out_data);
+
+    h.single_task<Consumer>([=]() [[intel::kernel_args_restrict]] {
+      for (size_t i = 0; i < size; i++) {
+        const T data_to_host_memory = ConsumePipe<T>::read();
+        *(h_out_data + i) = data_to_host_memory;
```
pmpeter1, line 114: why not h_out_data[i] = data_to_host_memory;   ?

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/zero_copy_data_transfer/src/zero_copy_data_transfer.cpp
```diff
@@ -0,0 +1,184 @@
...
+  size_t size = 10000;
+  size_t iterations = 1;
+#else
+  size_t size = 100000000;
+  size_t iterations = 5;
+#endif
+
+  // Allow the size to be changed by a command line argument
+  if (argc > 1) {
+    size = std::stoi(std::string(argv[1]));
```
pmpeter1, line 38: stoi might throw an exception.

pmpeter1 (changes requested):

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/zero_copy_data_transfer/src/restricted_usm_kernel.hpp
tyoungsc: host_ptr operator[] has issues.

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/zero_copy_data_transfer/src/restricted_usm_kernel.hpp
tyoungsc: host_ptr operator[] has issues.

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/zero_copy_data_transfer/src/restricted_usm_kernel.hpp
```diff
@@ -0,0 +1,142 @@
...
+using ConsumePipe = pipe<class ConsumePipeClass, T, kPipeDepth>;
+
+//
+// reads the input data from the hosts memory
+// and writes it to the ProducePipe
+//
+template <typename T>
+event SubmitProducer(queue& q, T* in_data, size_t size) {
+  auto e = q.submit([&](handler& h) {
+    h.single_task<Producer>([=]() [[intel::kernel_args_restrict]] {
```
keryell, line 64:
```suggestion
  return q.single_task<Producer>([=]() [[intel::kernel_args_restrict]] {
```

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/zero_copy_data_transfer/src/restricted_usm_kernel.hpp
```diff
@@ -0,0 +1,142 @@
...
+//
+// The worker kernel in the device:
+//  1) read input data from the ProducePipe
+//  2) perform computation
+//  3) write the output data to the ConsumePipe
+//
+template <typename T>
+event SubmitWorker(queue& q, size_t size) {
+  auto e = q.submit([&](handler& h) {
+    h.single_task<RestrictedUSM>([=]() [[intel::kernel_args_restrict]] {
```
keryell, line 88:
```suggestion
  return q.single_task<RestrictedUSM>([=]() [[intel::kernel_args_restrict]] {
```

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/zero_copy_data_transfer/src/restricted_usm_kernel.hpp
```diff
@@ -0,0 +1,142 @@
...
+      for (size_t i = 0; i < size; i++) {
+        const T data = ProducePipe<T>::read();
+        const T value = data * i; // perform computation
```
keryell, line 91: Is there some interest in specifying `const` here?
```suggestion
        auto data = ProducePipe<T>::read();
        auto value = data * i; // perform computation
```

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/zero_copy_data_transfer/src/buffer_kernel.hpp
```diff
@@ -0,0 +1,55 @@
...
+  // start timer
+  auto start = high_resolution_clock::now();
+
+  {
+    // set up the input/output buffers
+    buffer in_buf(in);
+    buffer out_buf(out);
+
+    // launch the computation kernel
+    auto kernel_event = q.submit([&](handler& h) {
```
keryell, line 29:
```suggestion
    q.submit([&](handler& h) {
```

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/zero_copy_data_transfer/src/buffer_kernel.hpp
```diff
@@ -0,0 +1,55 @@
...
+template <typename T>
+double BufferKernel(queue& q, std::vector<T>& in, std::vector<T>& out,
+                    const unsigned int size) {
+  // start timer
+  auto start = high_resolution_clock::now();
+
+  {
+    // set up the input/output buffers
+    buffer in_buf(in);
+    buffer out_buf(out);
```
keryell, line 26: This example is a little bit misleading because since you are using a buffers you do not really need vectors at the first place.
So it would be interesting to see what is the performance by using just buffers and using some host accessor when you need to initialize or access the data from the host.

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/zero_copy_data_transfer/src/buffer_kernel.hpp
```diff
@@ -0,0 +1,55 @@
...
+  // which will wait until the kernels finish and copy the data back to the
+  // host (if the buffer was written to).
+  // Therefore, at this point in the code, we know the kernels have finished
+  // and the data has been transferred back to the host.
+
+  // stop the timer
+  auto end = high_resolution_clock::now();
+  duration<double, std::milli> diff = end - start;
+
+  return diff.count();
```
keryell, line 52:
```suggestion
  return duration<double, std::milli> { high_resolution_clock::now() - start }.count();
```

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/zero_copy_data_transfer/src/restricted_usm_kernel.hpp
```diff
@@ -0,0 +1,142 @@
...
+// reads the input data from the hosts memory
+// and writes it to the ProducePipe
+//
+template <typename T>
+event SubmitProducer(queue& q, T* in_data, size_t size) {
+  auto e = q.submit([&](handler& h) {
+    h.single_task<Producer>([=]() [[intel::kernel_args_restrict]] {
+      // using a host_ptr tells the compiler that this pointer lives in the
+      // hosts address space
+      host_ptr<T> h_in_data(in_data);
```
keryell, line 67: Curious, in the other kernel this declaration is in the command group and not in the kernel.
Can it be anywhere?

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/zero_copy_data_transfer/src/restricted_usm_kernel.hpp
```diff
@@ -0,0 +1,142 @@
...
+  auto consumer_event = SubmitConsumer<T>(q, out, size);
+
+  // wait for all the kernels to finish
+  q.wait();
+
+  // stop the timer
+  auto end = high_resolution_clock::now();
+  duration<double, std::milli> diff = end - start;
+
+  return diff.count();
```
keryell, line 139: Fuse in 1 line?

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/zero_copy_data_transfer/src/zero_copy_data_transfer.cpp
```diff
@@ -0,0 +1,184 @@
...
+
+    // The FPGA emulator does not accurately represent the hardware performance
+    // so we don't print performance results when running with the emulator
+#ifndef FPGA_EMULATOR
+    // Compute the average latency across all iterations.
+    // We use the first iteration as a 'warmup' for the FPGA,
+    // so we ignore its results.
+    double buffer_avg_lat = std::accumulate(buffer_kernel_latency.begin() + 1,
+                                            buffer_kernel_latency.end(), 0.0) /
+                            (double)(iterations - 1);
```
keryell, line 151:
```suggestion
                            (iterations - 1);
```
already a double because of 0.0

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/zero_copy_data_transfer/src/zero_copy_data_transfer.cpp
```diff
@@ -0,0 +1,184 @@
...
+    double restricted_usm_avg_lat =
+        std::accumulate(restricted_usm_latency.begin() + 1,
+                        restricted_usm_latency.end(), 0.0) /
+        (double)(iterations - 1);
```
keryell, line 155:
```suggestion
        (iterations - 1);
```

keryell (changes requested):
Interesting example.
I am curious about how it works today.
Does `malloc_host` allocate physical contiguous memory?
Does the transfer involve some IOMMU to do the virtual address translation or all the transfers are using physical addresses?

A few possible simplifications otherwise.

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/zero_copy_data_transfer/src/restricted_usm_kernel.hpp
tyoungsc: I realize this was not very descriptive. Operator[] returns a const value, so using it gives this error for writing:

`error: cannot assign to return value because function 'operator[]' returns a const value h_out_data[i] = data_to_host_memory;`

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/zero_copy_data_transfer/src/restricted_usm_kernel.hpp
tyoungsc: This is a great change, but this would be the only tutorial using this format. So, while I will note it, I don't want to break the mold with this tutorial.

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/zero_copy_data_transfer/src/restricted_usm_kernel.hpp
tyoungsc: Same

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/zero_copy_data_transfer/src/buffer_kernel.hpp
tyoungsc: Keeping for consistency

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/zero_copy_data_transfer/src/restricted_usm_kernel.hpp
tyoungsc: I like being explicit here. Just my opinion :)

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/zero_copy_data_transfer/src/buffer_kernel.hpp
tyoungsc: I agree we don't need vectors AND buffers here, it was just to make it more like our other tutorials.

You don't need vectors, but you still need to copy those buffers from host to device. The fact that we are avoiding ANY copy to/from DDR by using host pointers allows us to avoid that overhead. Using host accessors would just make the memory movement more explicit, no?

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/zero_copy_data_transfer/src/buffer_kernel.hpp
tyoungsc: I agree. I proposed this originally, but I was ignored. I also proposed a helper function to grab sycl profiling info.

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/zero_copy_data_transfer/src/buffer_kernel.hpp
keryell: You can control carefully the memory transfers by picking the right model of accessors and not creating buffers from host data when useless. For example why initializing `out_buf` from `out`? But of course it changes a little bit the code organization.

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/zero_copy_data_transfer/src/restricted_usm_kernel.hpp
keryell: Ah sorry, I had a misconception about SYCL. I thought it was about simplifying FPGA programming compared to HLS, OpenCL or RTL... ;-)

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/zero_copy_data_transfer/src/buffer_kernel.hpp
tyoungsc: I think there are a couple things here. I will try to break them down.

> why initializing out_buf from out?

So that, when the kernel finishes, the data is automatically gets copied from out_buf to out. The user shouldn't have to explicitly copy to a host pointer or use a host accessor.

> Doesn't that incur a useless copy from out to out_buf before starting the kernel?

Yes. However, we think that is a bug and are trying to fix that in SYCL.

> Can the user not get better performance by explicitly moving memory?

Yes, and that was the point of the explicit_memory_movement tutorial. However, we don't think this should be a users (especially a new users) default. They should start with implicit data movement and change to explicit when they are 1) happy with the functionality and 2) think they get get performance increases.

Lastly, it is worth mentioning, even if the buffers were handled perfectly in this case, the zero-copy data transfer will still perform better since it avoids ANY overhead in copying to/from DDR and does everything over PCIe.

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/zero_copy_data_transfer/src/restricted_usm_kernel.hpp
tyoungsc: And  more!
However, there is pro for unity across tutorials as well. If deemed apprpriate, we would make this change across all samples at the same time.

tomlenth (dismissed):
No comments. looks good to me.  thank you

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/zero_copy_data_transfer/src/restricted_usm_kernel.hpp
tyoungsc: Currently, it must be within the `submit` block. We are working to make it work such that it can be declared anywhere :) We need to make sure the metadata stays connected to it though!

akertesz (approved):

tyoungsc:
@pmpeter1 could you please review my changes and let me know if things look good ?

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/zero_copy_data_transfer/src/buffer_kernel.hpp
pmpeter1: You can make a pull request against the dpc_common.hpp file to propose those changes.

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/zero_copy_data_transfer/src/zero_copy_data_transfer.cpp
```diff
@@ -0,0 +1,184 @@
...
+
+    // the golden output
+    std::vector<Type> out_gold(size);
+
+    // input and output data for the buffer version
+    std::vector<Type> in_buffer(size), out_buffer(size);
+
+    // input and output data for the restricted USM version
+    // malloc_host allocates memory specifically in the host's address space
+    Type* in_restricted_usm = malloc_host<Type>(size, q.get_context());
```
pmpeter1, line 77: Similar comment on malloc_host.  Why ctx instead of just q?

For sycl::free you don't need the prefix since you are not using it elsewhere.

tyoungsc:
@pmpeter1  same thing here. Made those changes and added a comment about `sycl::free` being used.

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/zero_copy_data_transfer/src/zero_copy_data_transfer.cpp
```diff
@@ -0,0 +1,185 @@
...
+                    (iterations - 1);
+
+    std::cout << "Average latency for the buffer kernel: " << buffer_avg_lat
+              << " ms\n";
+    std::cout << "Average latency for the restricted USM kernel: "
+              << restricted_usm_avg_lat << " ms\n";
+#endif
+
+    // free the allocated host usm memory
+    // note that these are calls to sycl::free()
```
pmpeter1, line 164: The extra comment note is unnecessary

pmpeter1 (approved):

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/zero_copy_data_transfer/src/zero_copy_data_transfer.cpp
tyoungsc: I disagree.
I think reminding the user that this is not the standard C++ `free()` call is, at least, a nice reminder. If the argument is that they should be able to notice that there is an extra argument (`q`) to the `free` and infer that means `sycl::free`, then we go down a road that leads to us having no comments, ever.

akertesz (approved):

akertesz:
The test failures are expected due to beta09/10 syntax differences

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/zero_copy_data_transfer/src/zero_copy_data_transfer.cpp
```diff
@@ -0,0 +1,185 @@
...
+    free(in_restricted_usm, q);
+    free(out_restricted_usm, q);
```
keryell, line 166: Just for fun we should add to the SYCL spec this shortcut `sycl::free(void *, sycl::queue q = {})` so you can afterwards write this really confusing coding style like `free(out_restricted_usm);` when not using `sycl::` ;-)
We are reaching the limits here... :-)  @jbrodman @gmlueck
On the obfuscation side, we have also `copy`, `memcpy` and all the group algorithms to write funny tutorials...

PR #256, opened 2020-10-16
GZIP update seed and remove USM flag

anandhv:
Update gzip, change seed and remove USM flag.


akertesz (approved):

akertesz (approved):

akertesz:
The test failures are expected failures:
 - Linux fails because the test machine is Ubuntu 20.4, which is not yet supported for FPGA compile
 - Windows fails because the test machine runs beta09, but GZIP requires features new to beta10

PR #258, opened 2020-10-16
Add 'all' target to oneMKL samples for Eclipse

petercad:
# Description

Adds "all" targets for oneMKL sample makefiles which do not have them. This is the default target for Eclipse.

## Type of change


anjgola:
These test failures are expected on Beta09. Merging the rest of the fix.

PR #275, opened 2020-10-26
ApplicationDebugger/array-transform: use STATUS kind in CMake message

barisaktemur:
# Description

Use the STATUS kind instead of WARNING for the messages printed from CMake.


JoeOster (approved):

akertesz (approved):

JoeOster:
fpga errors are known with Ubuntu 20, going to merge and see how it does with the nightly build, I expect no issues

PR #278, opened 2020-10-27
Fix oneVPL sample warnings and build instructions

mav-intel:
## Type of change

# How Has This Been Tested?

- [X] Command Line


JoeOster:
the cmake errors are related to beta10 since, at the time of running, beta 9 is still being used to test production

JoeOster:
Per Marc, this should work with Gold release, merging so that we can scan with compiler flags test

JoeOster (commented):
approving for ci and compiler flag tests

JoeOster (approved):

PR #281, opened 2020-10-28
Update top-level README to add new FPGA code samples

akertesz:
Signed-off-by: Audrey Kertesz <audrey.kertesz@intel.com>

# Description

Six FPGA code samples have been added for the 2021.1 oneAPI production release ("gold"). One FPGA code sample has changed its name. I'm updating the top-level README's table to reflect these changes.

## Type of change

- [X] README update


JoeOster (approved):
The changes are all thoroughly tested by @akertesz group’s own test framework. Currently (10/26/2020), every FPGA test will fail CI with the intel/INTEL issue. These changes remove the beta09 backward compatibility with the Beta10 workaround.

Samples CI daily scans hasn’t updated to beta10 yet, so every test on both Linux and Windows will fail.

When we do update to beta10, the above errors should be resolved and can be seen in the current nightly builds

Linux will still fail due to Ubuntu 20.4 that the testing is currently on. This is a known issue and is documented in the readme. That failure signature looks like “can’t find aoc”.

PR #337, opened 2020-11-12
Removed instances of "beta" (case insensitive) from readme markdown files

jedxops:
Removed instances of "beta" (case insensitive) from readme markdown files


JoeOster (changes requested):
Please update your personal repo

JoeOster (approved):

PR #438, opened 2021-03-09
Update -mkl -> -qmkl compiler flag for 2021.2

petercad:
This MR updates the oneMKL samples for compatibility with **2021.2 only**. There is a breaking change in the compiler flags for linking oneMKL (was `-mkl`, now `-qmkl`) between 2021.1 and 2021.2. Hence, CI is expected to fail.

Also fixes some compile errors in `monte_carlo_{pi,european_opt}` due to a missing `#include`.

This MR is to be merged after 2021.2 as PR#439 has implemented a fix so that 2021.1 and .2 oneMKL compiler will work, this PR will deprecate that fix

### DIFF
--- a/Libraries/oneMKL/black_scholes/GNUmakefile
+++ b/Libraries/oneMKL/black_scholes/GNUmakefile
@@ -14,8 +14,8 @@ mkl_path   := $(MKL)
 acc        := $(strip $(ACC))
 headers    := $(wildcard *.hpp)
 
-cxx_flags  := -O3 -DMKL_ILP64 -I$(MKLROOT)/include -fno-sycl-early-optimizations
-ldxx_flags := -mkl -fsycl-device-code-split=per_kernel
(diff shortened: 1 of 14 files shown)


JoeOster (approved):
approving for ci to run

JoeOster (changes requested):
We need to fix the following:
Add a comment about 

- the changes between 2021.1 and 2021.2 within the readme
- add some logic to the makefile to output useful guidance if the link fails, something like: "Build was not successful. Make sure you are using oneAPI 2021.2 or newer and have initialized the oneAPI environment using setvars.sh."

JoeOster:
@petercad - please update the sample, as you have a bunch of conflicts

petercad:
@JoeOster Yes, these are conflicts against #439. Let me close this PR for now, and I will reopen it with necessary changes after 2021.2 is ready.

PR #439, opened 2021-03-09
oneMKL samples: replace -mkl flag with full link line

petercad:
2021.2 introduces a breaking change where the `-mkl` compiler flag is replaced with `-qmkl`. This MR removes usage of the `-mkl` flag so that oneMKL samples work with both 2021.1 and 2021.2. The long term solution is to use `-qmkl` instead, as in #438.

This MR is intended to be merged into master after 2021.2 is tagged, so that master is compatible with both 2021.1 and 2021.2, but the the 2021.2 tag shows the intended, simpler `-qmkl` usage.

Also fixes a compilation error in a few samples due to a missing header file that used to be indirectly included.

FYI: @JoeOster


JoeOster (approved):
approving for CI purposes

PR #441, opened 2021-03-10
Update oneVPL samples to support 2021.2.2

mav-intel:
This is coupled with 2021.2.2 component release.

# Description

Import lib name has been changed for oneVPL 2021.2.2.  This impacts solution files since they require explicit file names to find libraries as opposed to using package configuration methods.

Ensure implementation supports 2.0 API

The oneVPL dispatcher has been extended to support hardware implementations with API versions lower than 2.0. Samples now correctly follow the oneVPL specification and check for implementation capabilities explicitly instead of assuming the first implementation found supports API 2.0.

# How Has This Been Tested?

- [x] Visual Studio
- [x] CMake

### DIFF
--- a/Libraries/oneVPL/dpcpp-blur/src/dpcpp-blur.cpp
+++ b/Libraries/oneVPL/dpcpp-blur/src/dpcpp-blur.cpp
@@ -140,12 +140,13 @@ int main(int argc, char *argv[]) {
     mfxLoader loader                    = NULL;
     mfxConfig cfg                       = NULL;
     mfxVariant impl_value               = {};
+    mfxVariant filter_value             = {};
     mfxSession session                  = NULL;
     mfxU16 input_width                  = 0;
     mfxU16 input_height                 = 0;
     mfxU16 out_width                    = 0;
     mfxU16 out_height                   = 0;
-    mfxVideoParam vpp_params            = { 0 };
+    mfxVideoParam vpp_params            = {};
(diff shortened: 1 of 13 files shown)


JoeOster (changes requested):
Note: Per conversation with Marc, further changes are needed

Libraries/oneVPL/dpcpp-blur/src/dpcpp-blur.cpp
```diff
@@ -202,15 +205,42 @@ int main(int argc, char *argv[]) {
...
+        {
+            printf("Found ApiVersion: %hu.%hu  ", idesc->ApiVersion.Major,
+                   idesc->ApiVersion.Minor);
+            printf("   \t%s ", (idesc->Impl == MFX_IMPL_TYPE_SOFTWARE) ? "SW" : "HW");
+
+            if (idesc->ApiVersion.Major >= 2)
+            {
+                sts = MFXCreateSession(loader, implIdx, &session);
+                printf("  session created\n");
+                break;
```
jonrecker, line 235: Also need to call MFXDispReleaseImplDescription each time through the loop.

JoeOster:
currently running into launching CI, error message as follows:, Snapshot dependency canceled: ... pre-execution checks

 merging as I believe its fine and we need it merged for PV

JoeOster (approved):

PR #446, opened 2021-03-15
Stream: update CMake build

petercad:
Fixes build error in stream (CMake is not liking the `.cc` extension for some reason). Also removes an unused source file.


JoeOster (approved):
approving for ci to run

PR #458, opened 2021-03-22
Updated CCL configurations to use CPU and improved sample instructions 

shailensobhee:
# Description

In the `sample.json` file, adapted the `[env]` section so that the `setvars.sh` are sourced with the proper flags such so that horovod can successfully install using the oneCCL backend. 

Also improved the README so that the sample instruction flow is clearer.


JoeOster (approved):
approving for ci

PR #466, opened 2021-04-06
Fix Formating

karlqi2000:
# Description
Formating in this file saxpy_func_solution_length.cpp (meant to be shown in Jupyter Notebook as a solution to an exercise)
was causing problems for Jupyter to display.


JoeOster (approved):
approving for ci

JoeOster (approved):

PR #475, opened 2021-04-14
WIP: DO NOT MERGE! Create common include directory for header file sharing between FPGA samples

mdbtucker:
Note: This PR is for discussion purposes and is NOT to be merged.

# Description

The oneAPI FPGA team would like to create an 'include' folder.  There are two main drivers behind this:

1. We currently have multiple code samples which use the same file.  Currently each sample has to have its own copy of this file.  This leads to different version of the same file in multiple places, which is confusing to users.
2. We would like to create a library of header files that contain useful classes optimized for FPGA performance.  We would like a common place where end users can go to get the latest version of these files.

To further elaborate on (2), this include folder is meant to house files that can be directly used by developers in their own projects.  Backwards compatibility will be a desirable trait of any change, but not an absolute requirement (developers are always free to simply not upgrade to a newer version in the case of an API change that breaks functionality for them).  The Code Samples project seems the ideal place for such a library of common files to reside.  We assume that other (non-FPGA) groups will eventually want to create similar folders for their own use.

As an example for discussion, this PR pulls the 'pipe_array' files (pipe_array.hpp and pipe_array_internal.hpp) into the common include folder.  The CMake files for the two Tutorials which reference these files are updated to use the new directory location (by adding the new 'include' directory to be included at compile time).  The sample.json files for these two Tutorials are also updated to reference the include directory using the 'commonFolder' feature (more on that later).  Note that the two Reference Designs that use pipe_array have NOT been similarly updated (see section on future work below).

This change takes advantage of the commonFolder feature described in the sample json specification (https://github.com/oneapi-src/oneAPI-samples/wiki/sample-json-specification#common-shared-folder-sample).  It is my understanding that this feature has not yet been implemented in the CLI.  I believe adding support for that feature in CLI and the VSCode browser would be important.  Without that support, a user who uses the CLI to fetch an example design would also have to manually fetch the include folder.  Users who simply clone the entire repo will not have any problems.

# Future Work

- move all shared files from existing code samples that are thought to have broad utility to developers into the new include folder
- properly update the CMake files for all code samples to add the new include directory as needed
- update the sample.json files of all affected code samples to use the commonFolder feature
- implement and test the commonFolder feature in the CLI
- create a README for the include directory that describes each of the common header files

# Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Sample Migration (Moving sample from old repository after completing checklist established)

# How Has This Been Tested?

Command line testing of one of the Tutorials only, thorough testing has not been done as this PR is not intended to be merged.


mdbtucker:
Notifying @jessicadavies-intel and @tyoungsc of this PR

JoeOster:
I suggest instead of having a common folder for just fpga header files, add it to the current common folder: https://github.com/oneapi-src/oneAPI-samples/tree/master/common , perhaps in a folder called /fpga as we already have a folder ‘stb’ in there

mdbtucker:
> I suggest instead of having a common folder for just fpga header files, add it to the current common folder: https://github.com/oneapi-src/oneAPI-samples/tree/master/common , perhaps in a folder called /fpga as we already have a folder ‘stb’ in there

Thanks for the suggestion Joe.  We did discuss this option internally, I can't remember why we dismissed it.  @jessicadavies-intel  or @akertesz can you comment?

mdbtucker:
Hmm, I see the CI pipeline has failed.  Apparently it does not support the commonFolder feature (or supports in in a way different from how I interpreted it) and also does not sync the entire repository when running tests.  I will reach out to Alexey Koshokin to ask for assistance.

mdbtucker:
> > I suggest instead of having a common folder for just fpga header files, add it to the current common folder: https://github.com/oneapi-src/oneAPI-samples/tree/master/common , perhaps in a folder called /fpga as we already have a folder ‘stb’ in there
> 
> Thanks for the suggestion Joe. We did discuss this option internally, I can't remember why we dismissed it. @jessicadavies-intel or @akertesz can you comment?

(side comment - it's very annoying that this 'conversation' view does not seem to support threaded conversations for general comments?  am I missing something?)

I think I remember what the problem was with that common folder.  From the README in that folder: "These include files are installed locally with the product installation and can be located at %ONEAPI_ROOT%\dev-utilities\latest\include on your development system."  We do not want these FPGA common files to be shipped with the compiler.  Our thinking is that this puts a much higher bar on the requirement to remain backwards compatible and never introduce changes that change the API.  We want to maintain a bit more freedom with the files we plan to put in this header library.

@JoeOster does that logic make sense to you, and are you ok with the include directory staying where I have put it in this PR?

jessicadavies-intel:
> > I suggest instead of having a common folder for just fpga header files, add it to the current common folder: https://github.com/oneapi-src/oneAPI-samples/tree/master/common , perhaps in a folder called /fpga as we already have a folder ‘stb’ in there
> 
> Thanks for the suggestion Joe. We did discuss this option internally, I can't remember why we dismissed it. @jessicadavies-intel or @akertesz can you comment?

In addition to the backwards compatibility issue you mentioned Mike, we would like these common header files to be easy for anyone to create or update by simply opening a PR to the oneAPI-samples repo (i.e., the same process as creating a new code sample).

JoeOster:
> > > I suggest instead of having a common folder for just fpga header files, add it to the current common folder: https://github.com/oneapi-src/oneAPI-samples/tree/master/common , perhaps in a folder called /fpga as we already have a folder ‘stb’ in there
> > 
> > 
> > Thanks for the suggestion Joe. We did discuss this option internally, I can't remember why we dismissed it. @jessicadavies-intel or @akertesz can you comment?
> 
> (side comment - it's very annoying that this 'conversation' view does not seem to support threaded conversations for general comments? am I missing something?)
> 
> I think I remember what the problem was with that common folder. From the README in that folder: "These include files are installed locally with the product installation and can be located at %ONEAPI_ROOT%\dev-utilities\latest\include on your development system." We do not want these FPGA common files to be shipped with the compiler. Our thinking is that this would put a much higher bar on the requirement to remain backwards compatible and never introduce changes that change the API. We want to maintain a bit more freedom with the files we plan to put in this header library.
> 
> @JoeOster does that logic make sense to you, and are you ok with the include directory staying where I have put it in this PR?

That component and include directory are not dependent on the compiler and is not shipped with the compiler.

JoeOster:
> > > I suggest instead of having a common folder for just fpga header files, add it to the current common folder: https://github.com/oneapi-src/oneAPI-samples/tree/master/common , perhaps in a folder called /fpga as we already have a folder ‘stb’ in there
> > 
> > 
> > Thanks for the suggestion Joe. We did discuss this option internally, I can't remember why we dismissed it. @jessicadavies-intel or @akertesz can you comment?
> 
> In addition to the backwards compatibility issue you mentioned Mike, we would like these common header files to be easy for anyone to create or update by simply opening a PR to the oneAPI-samples repo (i.e., the same process as creating a new code sample).

Using the current common folder is not an issue when doing a PR, just include the hpp file in that folder

mdbtucker:
Joe suggested a meeting to discuss, I think that's the best approach I will set something up for next week.

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/sample.json
```diff
@@ -8,6 +8,13 @@
   "targetDevice": ["FPGA"],
   "builder": ["ide", "cmake"],
   "languages": [{"cpp":{}}],
+  "commonFolder": {
+    "base": "../../..",
+    "include": [
```
jessicadavies-intel, line 13: Do you think the list under "include" should also contain "Tutorials/DesignPatterns/compute_units"?

I think this is suggested by the following documentation (https://github.com/oneapi-src/oneAPI-samples/wiki/sample-json-specification):

`
"include": [
      "sample-folder-name",
`

> if the "commonFolder" object is present, it is expected to reference at least one folder, the sample folder

mdbtucker: Thanks Jessica, I agree.  I have added the compute units folder as you suggested.

jessicadavies-intel:
The samples-aggregator and oneapi-cli tools were verified to work as-is, with this usage of commonFolder.

mdbtucker:
Confirmed that the CI now supports include files using the commonFolder feature in the sample.json file!  Failures in mvdr are unrelated and understood.  Failures in gzip are due to failure to update the cmake files.  Closing this PR as it has now served its purpose.

PR #482, opened 2021-04-22
Change intel::ii attribute to intel::initiation_interval in all FPGA code samples

ajaykumarkannan:
## Description
Change [[intel::ii]] to [[intel::initiation_interval]] since the former syntax is deprecated and produces a warning as of (upcoming) release 2021.3. 

## External Dependencies
None

## Type of change

## How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration: 

Ran the original and changed example designs with 2021.3. The original versions will emit a deprecation warning for ii, whereas 2021.3 will accept them without any warnings. I observed the impact of this attribute and verified that the compiler is behaving in the way that is expected (validated using the HTML reports). 

- [x] Command Line
- [x] oneapi-cli
- [ ] Visual Studio
- [ ] Eclipse IDE
- [ ] VSCode

### DIFF
--- a/DirectProgramming/DPC++FPGA/ReferenceDesigns/db/src/db_utils/fifo_sort.hpp
+++ b/DirectProgramming/DPC++FPGA/ReferenceDesigns/db/src/db_utils/fifo_sort.hpp
@@ -769,7 +769,7 @@ void sort(Compare compare) {
   constexpr int kTotalIter = kOutputStartLastStage + sort_size;
 
   // Sort
-  //[[intel::ii(1)]]
+  //[[intel::initiation_interval(1)]]
   for (int i = 0; i < kTotalIter; i++) {
 #ifdef DEBUG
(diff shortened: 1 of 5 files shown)


ajaykumarkannan:
@mdbtucker

JoeOster:
I found two places in the readme using ii, but not positive that these abbreviations shouldn't be changed as well

mdbtucker:
> I found two places in the readme using ii, but not positive that these abbreviations shouldn't be changed as well

Joe, I can't find any instances of 'ii' in any of the README files, can you point me to the ones you saw?

ajaykumarkannan:
Updated testing. Just checked; there shouldn't be "intel::ii" in any of our READMEs. John added a new tutorial with this attribute yesterday, so that is already using the new syntax.

PR #484, opened 2021-04-27
Changed FakeIOPipes for MVDR and IO streaming samples

tyoungsc:

## Description
FakeIOPipes no return a pair of events: the DMA event (a noop for USM host allocations) and the
kernel event. This allows the user of the FakeIOPipe to be more fine
grain on what they wait for. Used this in MVDR to not include the DMA
events in the throughput calculation for non-USM boards.

## External Dependencies

None

## Type of change

## How Has This Been Tested?
- [X] Command Line
Tested both changed tests in emulation for USM and non-USM boards.

### DIFF
--- a/DirectProgramming/DPC++FPGA/ReferenceDesigns/mvdr_beamforming/src/FakeIOPipes.hpp
+++ b/DirectProgramming/DPC++FPGA/ReferenceDesigns/mvdr_beamforming/src/FakeIOPipes.hpp
@@ -3,6 +3,7 @@
 
 #include <iostream>
 #include <type_traits>
+#include <utility>
 
 #include <CL/sycl.hpp>
 #include <CL/sycl/INTEL/fpga_extensions.hpp>
@@ -149,7 +150,8 @@ class ProducerImpl : public ProducerConsumerBaseImpl<Id, T, use_host_alloc> {
   using Pipe = sycl::INTEL::pipe<PipeID, T, min_capacity>;
 
   // the implementation of the static
-  static event Start(queue &q, size_t count = BaseImpl::count_) {
+  static std::pair<event, event> Start(queue &q,
+                                       size_t count = BaseImpl::count_) {
     // make sure initialized has been called
     BaseImpl::initialized_check();
 
@@ -188,7 +190,7 @@ class ProducerImpl : public ProducerConsumerBaseImpl<Id, T, use_host_alloc> {
       });
     });
 
-    return kernel_event;
+    return std::make_pair(dma_event, kernel_event);
   }
 };
 ////////////////////////////////////////////////////////////////////////////////
@@ -217,7 +219,8 @@ class ConsumerImpl : public ProducerConsumerBaseImpl<Id, T, use_host_alloc> {
   // the pipe to connect to in device code
   using Pipe = sycl::INTEL::pipe<PipeID, T, min_capacity>;
 
-  static event Start(queue &q, size_t count = BaseImpl::count_) {
+  static std::pair<event, event> Start(queue &q,
+                                       size_t count = BaseImpl::count_) {
     // make sure initialized has been called
     BaseImpl::initialized_check();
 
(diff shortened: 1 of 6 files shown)


DirectProgramming/DPC++FPGA/ReferenceDesigns/mvdr_beamforming/src/FakeIOPipes.hpp
```diff
@@ -245,21 +248,18 @@ class ConsumerImpl : public ProducerConsumerBaseImpl<Id, T, use_host_alloc> {
     });
 
     // if the user wanted to use board memory, copy the data back to the host
+    event dma_event;
```
mdbtucker, line 251: Should we initialized dma_event?  Or does it have a constructor that sets it to a reasonable default?

DirectProgramming/DPC++FPGA/ReferenceDesigns/mvdr_beamforming/src/mvdr_beamforming.cpp
```diff
@@ -305,7 +305,11 @@ int main(int argc, char *argv[]) {
     // By waiting on this kernel to finish, we are assuring that the runtime
     // has already programmed the FPGA with the image for this kernel. This
     // makes calling SetupPAC below safe (for the real IO pipes).
-    SinThetaProducer::Start(q, kNumSteer).wait();
+    event steer_dma_event, steer_kernel_event;
+    std::tie(steer_dma_event, steer_kernel_event) =
+      SinThetaProducer::Start(q, kNumSteer);
+    steer_dma_event.wait();
+    steer_kernel_event.wait();
```
mdbtucker, line 312: Just for my own understanding - the kernel event already depends on the dma event, so it would be equally correct to just wait on the kernel event here, right?  I like this code anyway for clarity, just want to make sure I understand.

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/io_streaming/src/FakeIOPipes.hpp
```diff
@@ -3,6 +3,7 @@
```
mdbtucker, line 3: I assume this file is now identical to the copy in the mvdr design?  (If this is not true - why not?)

DirectProgramming/DPC++FPGA/ReferenceDesigns/mvdr_beamforming/src/FakeIOPipes.hpp
tyoungsc: It will be initialized to a reasonable default (think of it like a 'null' event). If we call `.wait()` on it, it does nothing.

DirectProgramming/DPC++FPGA/ReferenceDesigns/mvdr_beamforming/src/mvdr_beamforming.cpp
tyoungsc: This understanding is correct. I just wanted to be explicit.

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/io_streaming/src/FakeIOPipes.hpp
tyoungsc: Almost. The one here has a comment explaining the `detail` namespace. Otherwise they are identical.

tyoungsc: I didn't like my answer. They are now the same.

mdbtucker (approved):

tyoungsc:
Are we not using the most up-to-date SYCL version in the runner?

mdbtucker:
> Are we not using the most up-to-date SYCL version in the runner?

No, the CI checks use the latest released version.  This seems like a problem to me, I will bring it up with the Code Samples workgroup.  In the mean time, failed validation is the expected result, so submitting this change.

PR #491, opened 2021-05-03
Update compute_units to use Linux-style flags

jessicadavies-intel:
## Description

Update the compute_units FPGA tutorial to use Linux-style flags on Windows.

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

- [x] Command Line

### DIFF
--- a/DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/CMakeLists.txt
+++ b/DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/compute_units/src/CMakeLists.txt
@@ -18,7 +18,7 @@ endif()
 
 # This is a Windows-specific flag that enables exception handling in host code
 if(WIN32)
-    set(WIN_FLAG "/EHsc")
+    set(WIN_FLAG "-EHsc")
 endif()
 
 # A DPC++ ahead-of-time (AoT) compile processes the device code in two stages.


mdbtucker (approved):
Approving to trigger running the CI pipeline.

mdbtucker:
@JoeOster, @jessicadavies-intel  found an FPGA sample that uses a windows style command line switch.  Can you please confirm:
a) that this is the correct solution (change from /EHsc to -EHsc), I'm assuming ALL command line switches need to change even these windows-only ones?
b) whether this change needs to go in for 2021.3, or if it needs to wait until 2021.4, or if either is acceptable?  It's not clear to me when this change is going into the compiler.

mdtoguchi (commented):
`/EHsc` and `-EHsc` are functionally equivalent.   Using the Windows MSVC compatible driver allows for both.

mdbtucker:
> `/EHsc` and `-EHsc` are functionally equivalent. Using the Windows MSVC compatible driver allows for both.

Thanks Michael, but I'm afraid I'm still a bit confused about the requirements.  Here's my understanding, please confirm if this is correct:
 - Using linux-style options (- instead of /) will work on both Linux and Windows as long as you use dpcpp as your compiler executable
 - Using Windows-style options will work on Windows only if you use dpcpp-cl as your compiler executable
 - This change will take effect with oneAPI 2021.3 release

If that's all correct, then the simplest solution to me seems to be to switch all the Code Samples to use linux-style options in both linux and Windows, and not used the dpcpp-cl executable at all.  Does my proposed solution (which Jessica has implemented here) seem reasonable to you?  Are there any drawbacks to this plan (for example, is the plan to deprecate support for dpcpp in Windows and force all users to use dpcpp-cl in Windows at some future time)?

mdtoguchi:
@mdbtucker, your summary is correct.  I think the confusion is the definition of 'linux style options'.  The option specifier (`/` or `-`) is not enough to qualify an option that is specifically for Windows or Linux.  The option here of `/EHsc` is a Windows MSVC compatible option, which means it will only work for Windows.  `/EHsc` is not available on Linux, even if specified with `-EHsc`.  There are a number of options which fall into this category which is best seen when using `dpcpp-cl -help -v` and checking the section under `CL.EXE COMPATIBILITY OPTIONS`.

mdbtucker:
Just met with Michael Toguchi, it seems this change is neither necessary nor recommended.  Sorry for the false alarm Jessica!  I will provide more details shortly, but closing this PR for now.

PR #523, opened 2021-05-27
Small update to FPGA DB design

tyoungsc:
# Existing Sample Changes
## Description
Small change that forces II=1 to all main processing loops. This works around a regression from a recent xmain integration AND is more clear since we know we want II=1 on all of these loops.

## Type of change

## How Has This Been Tested?
- [X] Command Line


mdbtucker:
@tyoungsc as discussed offline, let's try to get the compiler change that makes this design change necessary fixed, rather than changing the design.  If that's not possible, I'll approve this PR for 2021.4.

pmpeter1 (approved):

JoeOster (approved):

tomlenth (approved):

mdbtucker (approved):
The compiler change to fix this regression did not make it in to 2021.4, so we will have to update the code sample for now.

PR #551, opened 2021-06-03
Update Readmes for TM and Reg consistency

JoeOster:
Updating readme files for consistency
(TM) and ™ = ™
(R) and ® = ®


DirectProgramming/DPC++/Jupyter/oneapi-essentials-training/06_Intel_VTune_Profiler/src/iso3dfd_kernels.cpp
```diff
@@ -1,421 +1,421 @@
-//==============================================================
```
mdbtucker, line 1: You slipped in a change to a .cpp file here, was that intentional?  The diff is not showing me anything useful, I can't tell what has changed?

mdbtucker:
I just noticed there are a lot of whitespace changes included in these files as well.  It seems like maybe you changed all the end of line characters?  Was that intentional?

PR #556, opened 2021-06-09
Update FPGA loop_ivdep tutorial to use the recommended best practice

rho180:
Update FPGA loop_ivdep tutorial to explicitly specify the array when using the ivdep attribute

Signed-off-by: horobert <horobert@intel.com>

# Existing Sample Changes
## Description

Updated FPGA loop_ivdep tutorial to use array specifier in ivdep attribute, which is the recommended practice. Also updated README.md to reflect this change and the express the recommendation of using array specifiers.

### DIFF
--- a/DirectProgramming/DPC++FPGA/Tutorials/Features/loop_ivdep/src/loop_ivdep.cpp
+++ b/DirectProgramming/DPC++FPGA/Tutorials/Features/loop_ivdep/src/loop_ivdep.cpp
@@ -54,7 +54,7 @@ void TransposeAndFold(const device_selector &selector,
         // location that are less than kRowLength iterations apart.
         // The ivdep here instructs the compiler that it can safely assume no
         // loop-carried dependencies over safe_len consecutive iterations.
-        [[intel::ivdep(safe_len)]]
+        [[intel::ivdep(temp_buffer, safe_len)]]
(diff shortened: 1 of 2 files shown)


mendell27 (approved):
LGTM

DirectProgramming/DPC++FPGA/Tutorials/Features/loop_ivdep/src/loop_ivdep.cpp
```diff
@@ -54,7 +54,7 @@ void TransposeAndFold(const device_selector &selector,
         // location that are less than kRowLength iterations apart.
         // The ivdep here instructs the compiler that it can safely assume no
         // loop-carried dependencies over safe_len consecutive iterations.
-        [[intel::ivdep(safe_len)]]
+        [[intel::ivdep(temp_buffer, safe_len)]]
```
mdbtucker, line 57: As we discussed, it would be really great to add a second kernel with identical code but WITHOUT the ivdep.  Then the readme could first talk about the II problem that results, then show how the ivdep fixes the II problem.

mdbtucker (commented):
Good change Rob, thanks for doing this.  Readme needs to be fixed up (seems like you've based this off an old version), and if you have time I have suggested further enhancements to the tutorial.

PR #558, opened 2021-06-10
Template update

JoeOster:
Updating the PR template to reflect better UX and dropping of Protex and IPLDT pre scans


mdbtucker (approved):
LGTM

PR #571, opened 2021-06-29
Updates to FPGA Reference Design mvdr_beamforming for 2021.4

mdbtucker:
# Existing Sample Changes
## Description

 - modified StreamingQRD kernel to achieve II(1) after a compiler change in 2021.4 caused two local memory loads to no longer be combined
 - cause host program to exit if input data file is not available (used to hang)
 - cleanup of CMakeLists.txt
 - add initiation_interval(1) directive to CalcWeights kernel, required to achieve II(1) on Agilex devices
 - update Transpose kernel to use simpler arrays so that the desired memory organization is inferred when the number of columns is not a power of 2

## Type of change

## How Has This Been Tested?

- [x] Command Line


JoeOster (approved):
Approved for ci

pmpeter1 (approved):

PR #602, opened 2021-08-03
Fixed for compile error in MVDR

tyoungsc:
# Existing Sample Changes
## Description
Fixed a compile issue with MVDR. The compiler is generating a header file that was searching the global namespace for a KernelID that was a class member of a class. I changed the code to just used the passed in Id. This both fixes the compile and produces much nicer reports since the amount of name mangling in the kernel name is reduced. I also removed the Id parameter from the `ProducerConsumerBaseImpl` because it was unused.

## Type of change

## How Has This Been Tested?
- [X] Command Line
I tested the functionality in emulation.


mdbtucker:
Windows failed due to known issue with path length.

PR #607, opened 2021-08-05
simple change, need to verify current cmake version is working

JoeOster:
# Existing Sample Changes
## Description
The current sample config has minor changes to force validation, these samples failed in a separate large PR changing the min version to 3.4. This PR is to figure out remaining ci issues


louie-tsai:
@JoeOster 
For oneCCL failure, could you change [sample.json Line 20](https://github.com/JoeOster/oneAPI-samples/blob/update-cmake-version/Libraries/oneCCL/oneCCL_Getting_Started/sample.json#L20) from "make" to "make sycl_allreduce_test"?
I only saw cpu build from your log, so we could try to build only the binary needed.

PR #610, opened 2021-08-05
Update non FPGA samples no_init

JoeOster:
# Existing Sample Changes
## Description

With oneAPI 2021.4 the argument for accessors was changed from 'noinit' to 'no_init'. The change was derived from a change between the SYCL 2020 provisional spec and that of the 2020Rev3 spec


pmpeter1 (approved):

JoeOster:
These will fail for production ci as the changes are for 2021.4, currently waiting for preproduction ci to be updated to the newest compiler and should resolve itself at that time

JoeOster:
Sample fails CI as it requires 2021.4 bits

PR #614, opened 2021-08-10
Add ID field back to ProducerConsumerBaseImpl in FakeIOPipes.hpp

mdbtucker:
## Description

In FakeIOPipes.hpp, when two Producers or Consumers that use the same type are used in the same scope, their underlying base classes alias to each other and cause a runtime error when trying to initialize the second instance.  That error does not occur in the two samples that use FakeIOPipes, but it does happen in a new design (not yet comitted) that uses them.  This bug fix is for anyone who wants to reuse the FakeIOPipes in their own designs.

Also converted all line endings to UNIX style, so recommend ignoring whitespace changes when reviewing.

## Type of change

## How Has This Been Tested?

- [x] Command Line


mdbtucker:
CI failures are caused by running with 2021.3 when the samples have already been updated to 2021.4 terminology changes, to avoid deprication warnings.

JoeOster (approved):

PR #615, opened 2021-08-11
Fix for FPGA io_streaming design breaking due to non-forward declarable kernel name

tyoungsc:
# Existing Sample Changes
## Description
SYCL kernel names need to be forward declarable. The kernels in this tutorial were being falsely compiled until a recent change. This PR fixes the compile error.

## Type of change

## How Has This Been Tested?
- [X] Command Line
Built emulation and reports.


keryell (commented):
I have the feeling that replacing the `struct` by a `namespace` would have allowed to keep the same hierarchy and naming.

tyoungsc:
> I have the feeling that replacing the `struct` by a `namespace` would have allowed to keep the same hierarchy and naming.

Yep! You're right, but I wanted to shorten the mangled names anyway (long path issues and what not)

PR #619, opened 2021-08-12
Fixing intermittent segfaults in gzip design

tyoungsc:
# Existing Sample Changes
## Description
We were overrunning reads to a pointer in one of the kernels that caused an intermittent segfault in emulation. In theory, this segfault could occur in hardware too, but didn't seem to happen. We also didn't see segfaults in emulation or hardware for regular gzip, but we were overrunning the pointer, so I fixed that too.

## Type of change

## How Has This Been Tested?

- [X] Command Line
- [X] Hardware


JoeOster (approved):

tyoungsc:
All errors are expected because we are using an old version of the compiler. The path `#include <sycl/ext/intel/fpga_extensions.hpp>` was changed.

PR #635, opened 2021-08-23
Removing unnecessary -Xs* options from VCXPROJ files.

tyoungsc:
# Existing Sample Changes
## Description
The VS .vcxproj file uses "AdditionalOptions" field for compile options, and "SpecifyDevCmplAdditionalOptions" for link options. However, there is no way to specify it to use options when compiling for emulation vs reports. This causes emulation compiles to fail since they are passed a "-Xs" argument that it doesn't understand. This change removes a lot of -Xs arguments from the vcxproj file that are not necessary.

## Type of change

## How Has This Been Tested?
- [X] Command Line


mdbtucker (approved):

mdbtucker:
CI failures are expected due to use of 2021.4 syntax in these samples.

PR #639, opened 2021-08-27
FPGA gzip area fix

tyoungsc:
# Existing Sample Changes
## Description
Change to GZIP design to guard against segfaults without adding predication to the kernel. This is done by adding padding to the allocated input buffers/USM pointers so that the overrun in the kernel is still within the allocated region. This allows us to remove the predication and reduce the number of LSUs. This change also includes a seed update.

## Type of change

## How Has This Been Tested?
- [X] Command Line
- [X] regtest infra (s10, s10 usm, a10)


mdbtucker (approved):

tyoungsc:
Failures expected due to changes to fpga_extensions.hpp path.

PR #643, opened 2021-08-31
Update to FPGA DB acceleration reference design

tyoungsc:
# Existing Sample Changes
## Description
The following list outlines the changes made to the DB design:

- Changed all JOINs (merge, duplicate merge, and map) to use only the streaming method with done signals. I.e., we don't use index tracking and rely on the producers sending 'done' signals through the pipes.
- Refactors all JOINs to be functions rather than classes (cleaner)
- Refactored Q9, Q11, and Q12 to use the new JOINs
- Changed all queries to use implicit data transfer with buffers; previously, we had to use explicit transfers to workaround a SYCL runtime bug, but that bug has been fixed.
- Added sufficient padding to the end of all table buffers to avoid the need for predication on the LSUs in all queries. Instead the predication is on the computation after doing the loads. This allows for LSUs to be statically coalesced and saves area/improves performance.
- Changed all "producers" of data (i.e., those kernels which read from global memory and write to a SYCL pipe) to use loop induction variables that increment by 1 (before, they were incrementing by more than one). This improves the area/performance of the loop.
- Moved to blocking pipe reads in some places to simplify the kernel logic.
- Updated CMake file to have default seeds for A10, S10, and Agilex for each query
- Removed some dead code and minor code cleanup

## Type of change

## How Has This Been Tested?
- [X] Command Line
- [X] Regtest infra (A10, S10, and Agilex)


mdbtucker (approved):

tyoungsc:
Issues with `#include <sycl/ext/intel/fpga_extensions.hpp>` are expected. report_q11 and report_q9 failing are because the they are building using the default board, pac_a10, which is not supported for Q9 nor Q11 (requires S10 for area). @mdbtucker  what should we do here? I can change the default board for Q9 and Q11, but the report compile takes ~3 hours.... do we want to use that much CI time? Will it timeout?

mdbtucker:
> Issues with `#include <sycl/ext/intel/fpga_extensions.hpp>` are expected. report_q11 and report_q9 failing are because the they are building using the default board, pac_a10, which is not supported for Q9 nor Q11 (requires S10 for area). @mdbtucker what should we do here? I can change the default board for Q9 and Q11, but the report compile takes ~3 hours.... do we want to use that much CI time? Will it timeout?

I think it's better to have a long compile time in the CI than to have a failure in the CI.  Please change the default board for these querries to one that is supported.

mdbtucker (approved):

PR #650, opened 2021-09-01
Remove the use_library tutorial

ajaykumarkannan:
## Description
Remove the tutorial use_library due to HLS/OCL library support being removed in oneAPI 2021.4. Also note that RTL libraries has an issue (with a known workaround) so that part of the flow is also removed, but we will likely have a KDB to showcase the workaround. The fix for this is expected in 2022.1. Please use 2022.3 if you either need HLS/OCL library support in SYCL, or if you need continued access to RTL libraries. 

## External Dependencies
oneAPI 2021.4

## Type of change
Removal of a tutorial that is not supported in 2021.4

## How Has This Been Tested?
n/a since the tutorial is being removed


JoeOster (approved):

ndevenish:
> Please use 2022.3 if you either need HLS/OCL library support in SYCL, or if you need continued access to RTL libraries.

I assume this was intended to read "2021.3", unless it's expected to come back again?

ajaykumarkannan:
> I assume this was intended to read "2021.3", unless it's expected to come back again?

Yes, indeed. Fixed this in my comment above. Thanks for catching this!

PR #655, opened 2021-09-08
Declare kernel name in global namespace and update deprecated syntax in ProjectTemplates samples.

mdbtucker:
## Description

The ProjectTemplates samples were failing at runtime with 2021.4 compiler.  Moved the declaration of the kernel name class to be global to fix the issue.

Also cleaned up depricated syntax for the FPGA versions of the templates.

## Type of change

## How Has This Been Tested?

 - ran cmake (when present) and make on all updated samples, compiles all passed with no syntax warnings in 2021.4
 - executed the emulator targets for the fpga samples, they now work

- [x] Command Line


JoeOster (approved):

mdbtucker:
Compile failures for FPGA target are expected, due to the CI running 2021.3 version of the compiler.

PR #662, opened 2021-09-13
Add #include<string> to any sample that uses std::stoi

JoeOster:
## Description

The std::stoi function is defined in the \<string\> header file.  Several samples that use std::stoi were not explicitly including \<string\>, and were instead relying on some oneAPI headers to include it.  This change adds an explicit #include \<string\> to any sample that uses std::stoi.

Fixes Issue# 
https://jira.devtools.intel.com/browse/DEV1A-3037

## Type of change

## How Has This Been Tested?

- [x] Command Line
- [ ] oneapi-cli
- [ ] Visual Studio
- [ ] Eclipse IDE
- [x] VSCode


pmpeter1 (dismissed):
Why did  DirectProgramming/DPC++/StructuredGrids/1d_HeatTransfer/src/1d_HeatTransfer.cpp remove algoritm when it added string?

Also be caseful about ordering of #include and black lines.   These should be consistent (alphabetical) and blank lines between sections of different types of headers.

JoeOster:
> Why did DirectProgramming/DPC++/StructuredGrids/1d_HeatTransfer/src/1d_HeatTransfer.cpp remove algoritm when it added string?
> 
> Also be caseful about ordering of #include and black lines. These should be consistent (alphabetical) and blank lines between sections of different types of headers.

I removed "algorithm" as it was listed twice. As to lines, I see a few blanks lines added on accident, I will remove those.

mdbtucker (approved):
Approved for FPGA samples.

mdbtucker:
FPGA failures are expected because the CI runs with 2021.3

PR #699, opened 2021-10-12
Fix all warnings in mvdr_beamforming FPGA Reference Design

mdbtucker:
## Description

 - replace std::isnaan with sycl::isnaan
 - replace sycl::ONEAPI::memory_order with sycl::memory_order
 - replace e.get_cl_code() with e.get_code().value()

After these changes, the fpga_emu target compiles with no warnings.  The report target still has two warnings about properties being applied to loops that only execute once, but those warnings are expected.

## Type of change

## How Has This Been Tested?

- [x] Command Line


mdbtucker:
CI failures are expected, due to CI using 2021.4 version of the compiler.  The updated syntax in this PR requires 2022.1.

PR #739, opened 2021-11-05
Tools/ApplicationDebugger/jacobi: rework the jacobi sample.

nsaiapova:
# Existing Sample Changes
## Description

Restructure the sample, so there are function for initialization, main computation, prepare for the next iteration.
Add error computation.
Add built-in plan for debugging session that user is expected to follow.
Add hints to the user on how bugs can be found.
Extend README and add comments to the code.

## Fixes Issue 

Improves user experience.

## External Dependencies

None

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

- [X] Command Line
- [ ] oneapi-cli
- [ ] Visual Studio
- [ ] Eclipse IDE
- [ ] VSCode
- [X] When compiling the compliler flag "-Wall -Wformat-security -Werror=format-security" was used

@barisaktemur @tomlenth @xmnboy

### DIFF
--- a/Tools/ApplicationDebugger/jacobi/README.md
+++ b/Tools/ApplicationDebugger/jacobi/README.md
@@ -9,7 +9,7 @@ sample and the
 for the debugger.
 
 This sample contains two versions of the same program: `jacobi-bugged` and
-`jacobi-fixed`. The former works correctly, but in the latter, several bugs were
+`jacobi-fixed`. The latter works correctly, but in the former several bugs are
 injected. You can try to find and fix them using the debugger.
 
 | Optimized for       | Description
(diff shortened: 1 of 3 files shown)


praveenkk123 (approved):
Approve for CI only

barisaktemur (approved):
LGTM. Only minor comments.

praveenkk123 (approved):
Approve for CI

tomlenth (changes requested):
This is really looking great.  A couple of comments: 

- What device is run by default? Can you make note of that? 
- The three setvars commands are rendering as a single line on my browser.
- For jacobi bugged output, you provide bug 1 and bug 3.  Is there output for bug 2, and if not, can you just make note of that since it looks like a mistake by omitting it.

nsaiapova:
Hi @tomlenth,

There is no device that is used by default. A user is always expected to provide the argument.

I provide Bug 1 and Bug 2 outputs because these are outputs of the unmodified jacobi-bugged program. The Bug 2 will become visible once Bug 1 is fixed. I can add an example output for Bug 2 as well, but to get into this state a user would need to modify the program.

I could also make all bugs visible by default, so on CPU (and FPGA) you will see Bug 1 & 2 at once, and on GPU I think you would see only Bugs 3 & 1. Bug 3 happens at the very first iteration and exits the while-loop. And Bug 1 is triggered when the resulting vector does not match the expectation. Bug 2 will be skipped as the number of iterations would be 1.

tomlenth:
OK, I remember now that the device is required.  we should state that up front then as the error can be obscured in some cases. It would great to include the output with Bug 2.  I'd just note that it only becomes available once bug 1 is resolved. I don't think it's necessary to always show them.

xmnboy (approved):
Thanks for the updates, @nsaiapova. Makes it easier to understand how the app works. Also, I appreciate the initial set of examples showing how to interact with the SIMD lanes.

nsaiapova:
Restructured the output part of the README: now the bug 2 also included.
Fixed the rendering for setvars.sh commands.

tomlenth (approved):

PR #841, opened 2022-02-03
Fixing functional issue in DB query 11

tyoungsc:
# Existing Sample Changes
## Description
This change addresses a functional bug found in DB/Q11 where ivdep was used improperly. We were using ivdep without a safelen (safe length), while the caching technique we were using meant that we needed a safelen equal to the size of the cache.

Fixing this uncovered some performance issue that were also fixed in this PR. The design is now functional and performing the same as it was before.

Additionally, I have added a line of output that prints the "queries per second"  to measure throughput.

## Type of change

## How Has This Been Tested?

- [X] Command Line
- [X] regtests (S10 and Agilex, A10 is N/A).

### DIFF
--- a/DirectProgramming/DPC++FPGA/ReferenceDesigns/db/src/query11/query11_kernel.cpp
+++ b/DirectProgramming/DPC++FPGA/ReferenceDesigns/db/src/query11/query11_kernel.cpp
@@ -132,9 +132,8 @@ bool SubmitQuery11(queue& q, Database& dbinfo, std::string& nation,
 
       // populate MapJoiner map
       // why a map? keys may not be sequential
-      [[intel::initiation_interval(1), intel::ivdep]]
+      [[intel::initiation_interval(1)]]
       for (size_t i = 0; i < s_rows; i++) {
-        // read in supplier and nation key
         // NOTE: based on TPCH docs, SUPPKEY is guaranteed to be unique
         // in the range [1:kSF*10000]
         DBIdentifier s_suppkey = i + 1;
--- a/DirectProgramming/DPC++FPGA/ReferenceDesigns/db/src/query11/query11_kernel.cpp
+++ b/DirectProgramming/DPC++FPGA/ReferenceDesigns/db/src/query11/query11_kernel.cpp
@@ -4,7 +4,7 @@
 
 #include "query11_kernel.hpp"
 #include "pipe_types.hpp"
-#include "../db_utils/Accumulator.hpp"
+#include "../db_utils/CachedMemory.hpp"
 #include "../db_utils/MapJoin.hpp"
 #include "../db_utils/Misc.hpp"
 #include "../db_utils/Tuple.hpp"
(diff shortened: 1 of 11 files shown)


mdbtucker (approved):

DirectProgramming/DPC++FPGA/ReferenceDesigns/db/src/query11/query11_kernel.cpp
```diff
@@ -4,12 +4,11 @@
 
 #include "query11_kernel.hpp"
 #include "pipe_types.hpp"
-#include "../db_utils/Accumulator.hpp"
```
mdbtucker, line 7: You no longer use Accumulator.hpp in the Q11, but this PR includes changes to that file (as well as to other files in the db_utils directory).  Did you test all the db querries after your changes?

mdbtucker (approved):
Re-approving, hoping that will trigger a CI run (https://jira.devtools.intel.com/projects/REIE/issues/REIE-1367)

DirectProgramming/DPC++FPGA/ReferenceDesigns/db/src/query11/query11_kernel.cpp
```diff
@@ -120,33 +119,36 @@ bool SubmitQuery11(queue& q, Database& dbinfo, std::string& nation,
...
+      unsigned char nation_key_map_data[kSupplierTableSize + 1];
+      bool nation_key_map_valid[kSupplierTableSize + 1];
+      for (int i = 0; i < kSupplierTableSize + 1; i++) {
+        nation_key_map_valid[i] = false;
+      }
 
       // populate MapJoiner map
       // why a map? keys may not be sequential
-      [[intel::initiation_interval(1)]]
+      [[intel::initiation_interval(1), intel::ivdep]]
```
mdbtucker, line 135: I think a comment here explaining why ivdep is safe is justified.

Also, can we apply the ivdep to just a single variable?  I thought blanket ivdep was discouraged.

mdbtucker (approved):

DirectProgramming/DPC++FPGA/ReferenceDesigns/db/src/query11/query11_kernel.cpp
tyoungsc: Yes.

DirectProgramming/DPC++FPGA/ReferenceDesigns/db/src/query11/query11_kernel.cpp
tyoungsc: > I think a comment here explaining why ivdep is safe is justified.
Agreed. I will do so.

>Also, can we apply the ivdep to just a single variable?
I need to apply it to two variables. Maybe I can try `[[intel::initiation_interval(1), intel::ivdep(var1), intel::ivdep(var2)]]` ?

tyoungsc: Actually. I just realized I don't even need ivdep since the supplier keys are sequential, I can use the loop index variable as the index into the map and the compiler can determine that there is no loop carried deps itself. I will remove them.

mdbtucker (dismissed):

mdbtucker (dismissed):

mdbtucker (approved):

PR #885, opened 2022-03-11
depfiles break the HTML-file for FPGA optimization report

xwuupb:
The depfiles may break the generated HTML-file for FPGA optimization report.
The added line here needs to be added for every cmake files in DPC++FPGA.

# Existing Sample Changes
## Description

Recent cmake versions (e.g. 3.20.x and 3.21.x are tested) add the generation of depfiles via `CMAKE_DEPFILE_FLAGS_${LANG}` in Makefile generator, which may break the generated HTML-file for FPGA optimization report. Unset this variable can fix the problem for the generated `report.html`.


yuguen-intel:
I'm going to close this PR as I don't see any use case for this change.
Feel free to re-open it with more details, and we may proceed to do the batch change on all samples.

PR #917, opened 2022-04-04
FPGA QRD update seeds and improve program duration computation

yuguen-intel:
# Existing Sample Changes
## Description

Updating seed for 2022.2 to meet performance target and improve the way the duration of the computation is computed.

## Type of change

## How Has This Been Tested?

Ran a seed sweep on A10 and S10 and picked seeds that gave more than 24k matrices/s on A10 and more than 7k matrices/s on S10.


tyoungsc (approved):
Changes are good. Seed change is small and the other change is to measure the correct timing - no functional changes.

PR #934, opened 2022-05-10
Create shared header cached_local_memory.hpp and use it in onchip_memory_cache tutorial

mdbtucker:
This is just a preliminary review to get initial feedback.  I have more testing to do, and the readme needs to be cleaned up.  Just hoping to get some initial feedback.

There is a bit of 'ugliness' in the CachedLocalMemory class, which is that its data_ member variable needs to be public (even though it should never be accessed directly), and the loop that uses the cached memory needs to apply an ivdep that references that data_ member variable.  Still, I think this pattern comes up often enough that it justifies encapsulating into a shared header, even if the abstraction is not perfect.


mdbtucker:
@anandhv @yuguen could you both take a look at this?

DirectProgramming/DPC++FPGA/include/cached_local_memory.hpp
```diff
@@ -0,0 +1,123 @@
...
+    auto val = data_[addr];
+
+    // If this address was recently written, take the value from the cache
+    #pragma unroll
+    for (int i = 0; i < k_cache_depth; i++) {
+      if (cache_addr_[i] == addr) val = cache_val_[i];
+    }
+
+    return val;
+  }
```
yuguen-intel, line 63: I'm not sure to understand how this is working as a cache (but I can see that this was also what was done in the original cached_local_memory.hpp file).
When I think about a cache, I think about a "controler" that is the only interface with the memory. So no direct interaction with the memory is possible.
The cache will however manage:
- the writes to the memory if there is a write operation and it is full
- the reads from the memory if there is a read operation and the data is not present in the cache.

Here what is implemented is an history of the last writes - but I'm guessing that it also qualifies as a cache?

Now I'm (positively) surprised that our compiler is able to take advantage of this construct!
Can you confirm that the hardware generated is able to bypass the memory read (specially its latency) to read directly from the cache if the data is available?

DirectProgramming/DPC++FPGA/include/cached_local_memory.hpp
```diff
@@ -0,0 +1,123 @@
...
+#define __CACHED_LOCAL_MEMORY_HPP__
+
+#include <sycl/ext/intel/ac_types/ac_int.hpp>
+
+#include "constexpr_math.hpp"   // DirectProgramming/DPC++FPGA/include
+
+namespace fpga_tools {
+
+template <typename T,             // type to store in the memory
+          size_t k_mem_depth,     // depth of the memory
```
yuguen-intel, line 11: Looking at the other templated functions, I don't think that we prefix our integer template parameters with k_

DirectProgramming/DPC++FPGA/include/cached_local_memory.hpp
mdbtucker: The cache stores the most recent N writes to the memory.  If you read from the memory, and the read address matches an address in the cache, you pull data from the cache instead of from the memory itself.  That means as long as the read latency of the memory is <=N, you will get the correct value: values written more than N cycles ago can be safely fetched from the memory, and values written in the last N cycles are fetched from the cache, since they can't be read safely from the memory.  Does that make sense?

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/onchip_memory_cache/src/onchip_memory_cache.cpp
```diff
@@ -157,44 +120,42 @@ int main() {
...
+      t1_kernel = 
+        e.get_profiling_info<sycl::info::event_profiling::command_start>();
+      t2_kernel = 
+        e.get_profiling_info<sycl::info::event_profiling::command_end>();
       time_kernel = (t2_kernel - t1_kernel) / kNs;
 
       // Get accessor to output buffer. Accessing the buffer at this point in
       // the code will block on kernel completion.
-      host_accessor output_host(output_buf, read_only);
+      sycl::host_accessor output_host(output_buf);
```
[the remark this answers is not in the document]
mdbtucker: I write to the histogram memory a line or two later to clear it.  I noticed I had an error in my code that the test didn't catch, because the old (still correct) values were still sitting in the output buffer, so I added this to clear the output buffer.

DirectProgramming/DPC++FPGA/include/cached_local_memory.hpp
anandhv: And the memory read still happens -- but the result of the read is discarded if there's a cache hit.

DirectProgramming/DPC++FPGA/include/cached_local_memory.hpp
```diff
@@ -0,0 +1,123 @@
...
+          size_t k_cache_depth    // number of elements in the cache
```
anandhv, line 12: Apologies if this is explained elsewhere, but will you suggest a better way (compared to the previous version) to determined the value of k_cache_depth? Or will it still be trial-and-error?

DirectProgramming/DPC++FPGA/include/cached_local_memory.hpp
yuguen-intel: It makes perfect sense, I'm just happily surprised that the compiler does not insert pipeline registers after the cache read to match the memory read latency.

anandhv:
over seems good to me. The histogram code is so simple now...just read/write!

DirectProgramming/DPC++FPGA/include/cached_local_memory.hpp
mdbtucker: I really like this convention, I use it in all my templated code (including several places in the code samples).

kVariableName is the convention for constants.  Integer template parameters are compile time constants, but someone on the code samples group ages ago objected to using kVariableName format for them.  So this was my compromise to indicate that they are not 'variables', and can be used safely as constants, and will behave as constants in terms of performance.

You're in charge now, so if you want me to change this in the shared headers I will, but as I said I find this a helpful convention.

DirectProgramming/DPC++FPGA/include/cached_local_memory.hpp
mdbtucker: I'm going to run some experiments on that.  If the method currently described in the readme really is required (ie if read latency of 1 has fmax implications that do not show up in the reports) then that is a clear reporting bug, and I'll file a case to get it fixed.  If it doesn't have fmax implications, then there's no point talking about it, and all you need to do is achieve II1, which is much simpler.

DirectProgramming/DPC++FPGA/include/cached_local_memory.hpp
yuguen-intel: I think that this is a valid point and I'll make a note to update the other code samples to follow that convention

DirectProgramming/DPC++FPGA/Tutorials/DesignPatterns/onchip_memory_cache/src/onchip_memory_cache.cpp
```diff
@@ -7,138 +7,101 @@
...
-
-          // Cache is just a shift register, so shift the shift reg. Pushing
-          // into the back of the shift reg is done above.
-          #pragma unroll
-          for (int i = 0; i < kCacheDepth; i++) {
-            last_sum[i] = last_sum[i + 1];
-            last_sum_index[i] = last_sum_index[i + 1];
-          }
-        }
+      [[intel::ivdep(kSafeLen, histogram.data_)]]
```
whitepau, line 51: can the `ivdep` apply to `histogram` instead of `histogram.data_`?

DirectProgramming/DPC++FPGA/include/cached_local_memory.hpp
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  T read(addr_t addr) { 
+    auto val = data_[addr];
+    return val;
+  }
+
+  // data_ has to be public so we can use it in an ivdep attribute
+  // User code must NEVER modify data_ directly, it must be treated as private
+  T data_[k_mem_depth];
```
whitepau, line 111: why don't you mark this with the equivalent of `hls_memory` to force it into RAM (compiler can choose MLAB or M20K)?

DirectProgramming/DPC++FPGA/include/cached_local_memory.hpp
whitepau: Why can't the compiler figure this out? Realistically, you could create a cached RAM IP in RTL and this would happen transparently to the user (e.g. by applying an attribute such as `[[intel::memory_cached]]` to the array you want to have a cache on). The compiler can determine how deep to make the cache using static analysis at compile time.

whitepau (commented):
Jan and I wrote a similar class here for writing to off-chip memory:
https://github.com/intel-sandbox/whitepau.jan-resource-consumption/blob/trunk/externs.h

DirectProgramming/DPC++FPGA/include/cached_local_memory.hpp
mdbtucker: Paul what you're asking for is essentially for the compiler to modify the code at compile time.  It's not impossible, but also not a flow that's currently supported, and I don't think is likely to be supported.

DirectProgramming/DPC++FPGA/include/cached_local_memory.hpp
mdbtucker: It's not necessary.  If the compiler decides this should be implemented in registers instead of RAM, this class still works.

DirectProgramming/DPC++FPGA/include/cached_local_memory.hpp
whitepau: not at all: I'm asking for the compiler to use a RAM IP with a cache instead of whatever RAM IP we use by default. this won't affect the kernel behaviour at all since it's a local memory. This might be different with an off-chip memory because it would affect the side effects of the kernel to the outside world.

mdbtucker: Ah, I see what you're saying.  That's an excellent idea.  I was actually thinking about this as a way of working around the issues we discussed today related to ivdep.  I will be creating an internal case about the ivdep issues, I'll include the suggestion to have a cached RAM IP automatically used by the compiler, I think it will solve several issues and greatly simplify the user experience.

mdbtucker:
Closing this pull request, will create a new PR for the updated and very different code.

PR #964, opened 2022-06-14
Update FPGA onchip_memory_cache tutorial to create a reusable class and get rid of ivdep

mdbtucker:
## Description

This update hides the implementation of the cache in a reusable class (OnchipMemoryWithCache, in the shared include directory) and gets rid of the (previously improperly applied) ivdep attribute on the loop.

This change results in a small area increase (a cache depth of 7 is needed to achieve the same results as the original design with a cache depth of 5), but no performace (fmax or throughput) degradation.  The change also allows users to simply re-use the class in the shared header file, rather than having to implement it themselves, which is a process that was very prone to errors.  

More importantly, the previous version of this tutorial was using the ivdep attribute in an incorrect way that could result in functional errors.  Removing the ivdep entirely eliminates this possibility.

## External Dependencies

None

## Type of change

## How Has This Been Tested?

Ran seed sweeps with each relevant cache depth (0, 1, and 7) in isolation to verify no fmax drops from the previous version.

Ran the full design (hardware and emulation) on a10 and s10 devices in both Linux and Windows, all passed.


yuguen-intel:
Adding @tomlenth for the README changes and @pmpeter1 for the SYCL code changes.
(This PR has been reviewed internally before)

tomlenth (approved):
Looks good.  thanks!

pmpeter1 (approved):
Code style is good.   I see a an introduction of explicit sycl:: prefix.   I know this has been a discussion topic on coding guidelines, and the samples WG should have a deeper discussion on if we want to change this recommendation to be more explicit even though it's more verbose.

Please make sure all usages of the samples work with this approach.  This is both extracting one sample at a time via the sample-browser and checking out the entire repo.    Supporting both modes was why we ended up with the more complex solution for dpc_common.hpp

mdbtucker:
> Please make sure all usages of the samples work with this approach. 

I don't understand what you mean by 'this approach'?  If you're talking about the explicit use of sycl::, I don't see how that's related?  Perhaps you're referring to the shared include directory?  If so, we have many FPGA samples that use this shared directory, and it has been validated with the sample browser and when checking out the entire repo.

pmpeter1:
> 

The explicit use of sycl:: is a style issue.    If we want to move in this direction it means ratifying a change in the samples WG to the coding guidelines.    Given the community feedback this may be a good change.

By "this approach" I was referring to the use of the samples via the sample-browser.     If you have already validated that this code works with the sample-browser (all relative links are checked out correctly), then I have no additional concern.

yuguen-intel (approved):

yuguen-intel:
We've been trying to remove the use of `using namespace` in new FPGA samples.
I can bring this up in the next samples meeting.

PR #968, opened 2022-06-17
FPGA: Adding instructions on how to run the samples on Windows

yuguen-intel:
# Existing Sample Changes
## Description

This PR adds instructions in the FPGA samples READMEs to run the samples on Windows.

## Type of change


yuguen-intel:
Hey @tomlenth, could you have a look at these README changes?
These just add instructions on how to run the samples on Windows.
We historically avoided giving these instructions as we are not showing the samples outputs on Windows.
The product validation team asked to add these instructions as the README instructions are more generic than what we show.

tomlenth (changes requested):
In the other parts of the sample, Windows and Linux are split out into specific sections.  This is preferred as intertwining the two OS commands can be a little cumbersome.  I could be convinced otherwise, but in our user studies, placing the commands side by side seems to irritate some folks. Dedicated sections ensures a single path where the user doesn't have choose an OS at each step.

yuguen-intel:
Interesting
This is the case for most of the FPGA READMEs, except the run commands (e.g. to run the emulator, all the Linux and Windows commands are side by side), which is why I did the same thing as what was done already.
Do you think it is worth it that I go over all the READMEs and create a next section for that purpose?

tomlenth:
Well, the FPGA samples seem to kind of mix things.  For example, the Tutorial/Get Started/fast_recompile, The Build instructions are separated. But, the Run commands are combined.  The build commands are a bit more involved so definitely makes sense to separate.  The run section is pretty light.  And, Windows would only have one command if separated. maybe this is preferred.

yuguen-intel:
OK, I'll keep it as is for now then.
Reach out to me if you think that this needs to change.

jimmytwei:
@yuguen-intel Since these are README.md changes, I've requested @ishaghosh27 to review.

tomlenth (approved):

jimmytwei (approved):

PR #1011, opened 2022-07-21
Fix errors in histogram example

SergeyKopienko:
Fixing error: Assertion in histogram example occurred
Copied changes from
             https://github.com/oneapi-src/oneDPL/pull/573/files
into examples repository.


MikeDvorskiy (changes requested):
I believe that it should be absolutely the same to https://github.com/oneapi-src/oneDPL/tree/main/examples/histogram
In ideal we should have just reference to https://github.com/oneapi-src/oneDPL/tree/main/examples/histogram

MikeDvorskiy (approved):
LGTM

reble (approved):

jimmytwei (approved):

PR #1097, opened 2022-09-21
sample.json file was missing a GUID

xmnboy:
# Existing Sample Changes
## Description

sample.json file was missing a GUID

Fixes Issue# 

## External Dependencies

None

## Type of change


xmnboy:
BTW -- @praveenkk123 -- with this fix we need to also change the 2022.3.0 tag so it includes the fixes to the `sample.json` file in this sample. The fix is going to master, but the 2022.3.0 represents master just before this fix and the other fix was made to the `sample.json` file. I leave it to you to figure out the safest way to move the 2022.3.0 tag.

praveenkk123 (approved):

praveenkk123 (approved):

PR #1134, opened 2022-10-13
Initial rework of the vector-add sample

yuguen-intel:
# Existing Sample Changes
## Description

An initiative has been taken to rework the simple-add and vector-add samples.
This PR is a first step towards this overall changes.
It moves the project from using `make` to using `cmake`.
The readme and the `.sh` scripts have been updated accordingly.
The `sample.json` has also been updated, even though we don't really have a use for this file right now (and no way to test it).


yuguen-intel:
I've reverted my changes to the README as @jkinsky is going to make a PR for his own changes - both of our PRs should be merged at the same time.

jimmytwei (approved):

PR #1159, opened 2022-10-26
Add more comments in the vector-add CMakeLists.txt file

yuguen-intel:
This change adds instructions for the users to modify the CMakeLists.txt of the vector-add sample.
The idea is to make the vector-add sample as the empty template sample for new users.


xmnboy (approved):
The updated comments/instructions are clear to me -- and I am not an FPGA expert. :-)

jkinsky (approved):
Makes sense. The only thing to note is that instructions referencing "dpcpp" will have to be updated/changed when invocation changes in the next release. Those change based on the OS too.

jimmytwei:
I'll merge the change now, but like @jkinsky said, will need to update the driver based on invocation

PR #1186, opened 2022-11-03
[FPGA] Added sim target to gzip and printf

haoxian2:
# Existing Sample Changes
DirectProgramming/DPC++FPGA/ReferenceDesigns/gzip
DirectProgramming/DPC++FPGA/Tutorials/Features/printf

## Description

Add support for the FPGA simulator to run the code samples for gzip, gzip_ll, and printf.
Update README in to reflect not being able to printf **long int** in windows.
Added additional changes to ensure the buffer can be allocated accordingly and can be executed within a reasonable period of time, as simulator is 1000-10000x slower than hardware.

## Type of change

## How Has This Been Tested?

- [X] Command Line
- [ ] oneapi-cli
- [ ] Visual Studio
- [ ] Eclipse IDE
- [ ] VSCode
- [ ] When compiling the compiler flag "-Wall -Wformat-security -Werror=format-security" was used


DirectProgramming/DPC++FPGA/ReferenceDesigns/gzip/src/gzip.cpp
```diff
@@ -154,6 +156,10 @@ int main(int argc, char *argv[]) {
 
 #ifdef FPGA_EMULATOR
     CompressFile(q, infilename, outfilenames, 1, true);
+#elif FPGA_SIMULATOR
+    // warmup run
+    CompressFile(q, infilename, outfilenames, 1, false);
```
yuguen-intel, line 161: Why would you need a warmup run when running on the simulator?

DirectProgramming/DPC++FPGA/ReferenceDesigns/gzip/src/gzip_ll.cpp
```diff
@@ -160,6 +162,9 @@ int main(int argc, char *argv[]) {
 
 #ifdef FPGA_EMULATOR
     CompressFile(q, infilename, outfilenames, 10, true);
+#elif FPGA_SIMULATOR
+    CompressFile(q, infilename, outfilenames, 1, false);
```
yuguen-intel, line 166: I'm guessing that this is also a warmup run, but is this needed on the simulator?

DirectProgramming/DPC++FPGA/ReferenceDesigns/gzip/src/gzip.cpp
haoxian2: I would assume that simulator also has some first kernel setups similar to hardware. But I just realized that I'm also skipping the part where performance is measured in simulation, so regardless a warmup run is not needed :)

yuguen-intel (approved):

PR #1193, opened 2022-11-08
Fix simple-add sample.json and remove unused Makefile files

yuguen-intel:
This change modifies the simple-add `sample.json` file to reflect the new build cmake-based build commands.
It also removes the now unused Makefiles files.


jkinsky (approved):

jkinsky (approved):

jkinsky (changes requested):
Remove bullets or numbers from the second step, in both instances. Readers are smart enough to see the difference between two options without the order or unordered list format.

jkinsky (approved):

jimmytwei (approved):

PR #1277, opened 2023-01-05
removed dpc_common.hpp dependency

krisrak:
# Existing Sample Changes
## Description
removed dpc_common.hpp dependency from 2 sample codes that only affected CI testing

## Type of change

## How Has This Been Tested?
- [X] Command Line


krisrak:
will be submitting new PR for this since there is conflict after folder rename

PR #1372, opened 2023-02-22
FPGA: replace Agilex™ with Agilex®

yuguen-intel:


jkinsky (approved):
This is an accurate and required change according to the names database.

jimmytwei (approved):

PR #1656, opened 2023-05-12
FPGA Speculated Iterations sample readme update

jkinsky:
# Existing Sample Changes
## Description

Restructured to match new template-with adjustments for FPGA samples. Updated name in readme and sample.json file. Added category to readme. Corrected some markdown formatting. Updated some branding. Rewrote some text for clarity and to correct some grammar.

Fixes Issue# 

## External Dependencies

N/A

## Type of change

- [x] Documentation quality

## How Has This Been Tested?

- [x] Command Line
- [x] VSCode

### DIFF
--- a/DirectProgramming/C++SYCL_FPGA/Tutorials/Features/speculated_iterations/README.md
+++ b/DirectProgramming/C++SYCL_FPGA/Tutorials/Features/speculated_iterations/README.md
@@ -1,14 +1,27 @@
 
-# Speculated Iterations of a Loop
-This FPGA tutorial demonstrates applying the `speculated_iterations` attribute to a loop in a task kernel to enable more efficient loop pipelining.
+# `Speculated Iterations` Sample
 
-| Optimized for                     | Description
-|:---                               |:---
-| OS                                | Linux* Ubuntu* 18.04/20.04 <br> RHEL*/CentOS* 8 <br> SUSE* 15 <br> Windows* 10
-| Hardware                          | Intel® Agilex® 7, Arria® 10, and Stratix® 10 FPGAs
-| Software                          | Intel® oneAPI DPC++/C++ Compiler
-| What you will learn               | What the `speculated_iterations` attribute does <br> How to apply the `speculated_iterations` attribute to loops in your program <br> How to determine the optimal number of speculated iterations
-| Time to complete                  | 15 minutes
+This sample is an FPGA tutorial that demonstrates applying the `speculated_iterations` attribute to a loop in a task kernel to enable more efficient loop pipelining.
+
+| Area                 | Description
+|:---                  |:---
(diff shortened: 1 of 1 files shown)


DirectProgramming/C++SYCL_FPGA/Tutorials/Features/speculated_iterations/sample.json
```diff
@@ -1,6 +1,6 @@
 {
   "guid": "66A57127-1F8D-4769-8CCB-16ECC56A446F",
-  "name": "Speculated Iterations",
+  "name": "speculated_iterations",
```
yuguen-intel, line 3: Do we want to do that? This is the name that would appear in the code samples browser for this sample
1/ Do we want to keep plain English or the actual feature name?
2/ If we decide to make this change, we should make it to all the other samples as well (which I'm ok with as long as we keep them uniform

jkinsky: I was following the convention I thought was emerging from several other similar feature samples. I'm okay with whatever you decide.

yuguen-intel: I'd say:
- In the readme, we can use the name of the feature, e.g. `kernel_args_restrict` sample
- In the `samples.json`, use a human readable name as this is what people are going to browse, e.g. "Pointer Aliasing and the Restrict Directive" as this is the name of the doc page: https://www.intel.com/content/www/us/en/docs/oneapi/optimization-guide-gpu/2023-0/pointer-aliasing-and-the-restrict-directive.html

What are your thoughts? In that case you would not change the `sample.json` files and I would look for someone to do it

jkinsky: I think we should be consistent. If the readme has one name and the sample browsers use another (as determined by the name in the sample.json file) then it's confusing. We understand the difference. Others will not.

yuguen-intel: That makes sense, in that case what would be your recommendation:
1/ use the feature name, e.g. `kernel_args_restrict`
2/ use the doc name, e.g. "Pointer Aliasing and the Restrict Directive"

In any case, do not make the modifications to the sample.json files, I'll make it to all the samples at once.

jkinsky: I think we should use option 2, use the doc name. Make them readable to humans who are looking to solve problems. Using the feature name assumes a developer already knows what to look for.

yuguen-intel (approved):

PR #1661, opened 2023-05-15
Add an additional flag for the oneMKL Monte-Carlo Sample

andreyfe1:
# Existing Sample Changes
## Description

Add a flag to let users build the sample using oneMKL generation on host if needed (flag is switched off by default).

## Type of change

## How Has This Been Tested?

- [X] Command Line
- [x] oneapi-cli
- [ ] Visual Studio
- [ ] Eclipse IDE
- [ ] VSCode
- [X] When compiling the compliler flag "-Wall -Wformat-security -Werror=format-security" was used

Signed-off-by: Fedorov, Andrey <andrey.fedorov@intel.com>


IgorOchocki (changes requested):
This sample runs using double precision which gen 12 does not support. This sample will not work on some HW.

andreyfe1:
> This sample runs using double precision which gen 12 does not support. This sample will not work on some HW.

@IgorOchocki, correct. We need double precision for this sample. How can we say it in the Readme clearly? Do we have other samples that use double precision only? So, I could look at them and make corresponding changes to be aligned with them. Some our other samples use double precision as well

PR #1717, opened 2023-06-12
delete host memory usage which is allocated for device

ChengChen002:


jimmytwei:
@IgorOchocki Can you take a look at this?

IgorOchocki (approved):

PR #1792, opened 2023-07-11
Update fortran guided mat mul openmp

krisdale:
# Existing Sample Changes
## Description

Update 03_mm_GPU.f90 to resolve build error.

This addresses a reported non-zero error code. The update is correcting an inconsistency between source files with single vs double precision.

## Type of change

- [x ] Bug fix (non-breaking change which fixes an issue)


jimmytwei:
@krisdale Can you update this PR? It has a bunch of files that are not associated with the fix.

krisdale:
Abandon this PR as it is based on the wrong branch.

PR #1806, opened 2023-07-25
Make the design smaller to prevent timing violations

alisanikiforova:
# Existing Sample Changes
## Description

This test is periodically hits timing violations in S10 compiles, and this change makes the design smaller to make it more likely to fit without timing violations.

Fixes Issue#14019796464  

## Type of change

## How Has This Been Tested?

Tested using a regtest to confirm timing violations no longer come up in S10: https://spetc.intel.com/testsummary?trview=0&testRunIds=8426203
Ran the executable with the openCL Intercept Layer and confirmed the difference between where double buffering is and isn't applied is still visible.
(see image below - I used an older version of the intercept layer so it doesn't look the same as in the readme, but the difference in gaps between kernel executions is still visible)
image


whitepau (approved):
ok

PR #1836, opened 2023-08-16
FPGA: add a missing header in the ANR sample

broxigarchen:
# Existing Sample Changes
## Description

added missing header


yuguen-intel (commented):
Is this header really missing? It looks like this sample is passing all tests

broxigarchen:
> Is this header really missing? It looks like this sample is passing all tests

It seems there will be changes in new SYCL's header. This is a change to unblock the test as this change does no harm

yuguen-intel (approved):

PR #1894, opened 2023-09-15
Add flag to report for windows

dancgr:
We can't take out the Qactypes flag for the reports, as it is needed for windows report linking.


yuguen-intel:
Addressed in https://github.com/oneapi-src/oneAPI-samples/pull/1896 which added guards for the change to only occur when Windows was the target

PR #1950, opened 2023-10-06
fix: Make sure that VisualMandlebrot builds on linux

nedo99:
# Existing Sample Changes
## Description

* Add script that will download and install SDL2
* Update CI build steps

## External Dependencies

List any external dependencies created as a result of this change.

## Type of change

## How Has This Been Tested?

- [x] Command Line
- [ ] oneapi-cli
- [ ] Visual Studio
- [ ] Eclipse IDE
- [x] VSCode

### DIFF
--- a/DirectProgramming/C++SYCL/VisualizedSamples/VisualMandlebrot/get_sdl2.sh
+++ b/DirectProgramming/C++SYCL/VisualizedSamples/VisualMandlebrot/get_sdl2.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+git clone https://github.com/libsdl-org/SDL.git
+cd SDL
+
+git checkout release-2.0.20
+mkdir build
+mkdir install
+cd build
+cmake ../ -D CMAKE_INSTALL_PREFIX=../install
+make && make install
+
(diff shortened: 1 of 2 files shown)


DirectProgramming/C++SYCL/VisualizedSamples/VisualMandlebrot/get_sdl2.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+git clone https://github.com/libsdl-org/SDL.git
```
jimmytwei, line 3: @nedo99 Is there a reason you used the 2.0.20 version? Can we have the script use the latest version?
Something like this:

```
latest_version=`git ls-remote --tags https://github.com/libsdl-org/SDL.git | awk -F/ 'END{ print $3 }'`
git clone -b $latest_version https://github.com/libsdl-org/SDL.git
cd SDL
mkdir -p build
mkdir -p install
cd build
cmake ../ -D CMAKE_INSTALL_PREFIX=../install
make && make install
```

nedo99: We can of course have the latest version, but the idea here is to specify the version that we know works for sure. The risk with the latest SHA is that some changes might introduce some bugs or break our build. It is safer to specify the version and update it so we are in control.

PR #2189, opened 2024-02-16
FPGA: Added information about how to obtain BSP information in READMEs

yuguen-intel:
This change adds information about how to obtain BSP information in READMEs for guiding users in finding what arguments should be passed to `Xstarget=xxx`.

### DIFF
--- a/DirectProgramming/C++SYCL_FPGA/ReferenceDesigns/anr/README.md
+++ b/DirectProgramming/C++SYCL_FPGA/ReferenceDesigns/anr/README.md
@@ -253,10 +253,10 @@ The design uses the following generic header files.
       nmake fpga
       ```
 >**Note**: If you encounter any issues with long paths when compiling under Windows*, you may have to create your 'build' directory in a shorter path, for example `C:\samples\build`. You can then run cmake from that directory, and provide cmake with the full path to your sample directory, for example:
-
-  ```
+>
+>  ```
   > C:\samples\build> cmake -G "NMake Makefiles" C:\long\path\to\code\sample\CMakeLists.txt
-  ```
+>  ```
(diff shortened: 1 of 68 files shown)


DirectProgramming/C++SYCL_FPGA/ReferenceDesigns/anr/README.md
```diff
@@ -234,8 +252,11 @@ The design uses the following generic header files.
       ```
       nmake fpga
       ```
->**Note**: If you encounter any issues with long paths when compiling under Windows*, you may have to create your ‘build’ directory in a shorter path, for example `C:\samples\build`. You can then run cmake from that directory, and provide cmake with the full path to your sample directory.
+>**Note**: If you encounter any issues with long paths when compiling under Windows*, you may have to create your ‘build’ directory in a shorter path, for example `C:\samples\build`. You can then run cmake from that directory, and provide cmake with the full path to your sample directory, for example:
 
+  ```
+  > C:\samples\build> cmake -G "NMake Makefiles" C:\long\path\to\code\sample\CMakeLists.txt
+  ```
```
whitepau, line 259:
```suggestion
>**Note**: If you encounter any issues with long paths when compiling under Windows*, you may have to create your 'build' directory in a shorter path, for example `C:\samples\build`. You can then run cmake from that directory, and provide cmake with the full path to your sample directory, for example:
>
> ```
> C:\samples\build> cmake -G "NMake Makefiles" C:\long\path\to\code\sample\CMakeLists.txt
> ```
```

I might have gotten some of the syntax wrong, but the code snippet should appear inside the note box

Also, we shouldn't be using the weird fancy single quote marks

whitepau (changes requested):

PR #2193, opened 2024-02-20
FPGA: Remove fpga_register annotations from fifo_sort

justin-rosner:
## Description

A recent update to the SYCL front-end uncovered an issue in the compiler. The `fpga_register` attributes were never being applied to the member variables of the `Preloader` class. Instead, the compiler was using it's own heuristics to registerize the entire `Preloader` object (`preloadersA`). With the update, the compiler now sees the correct attributes to the member variables, but cannot break apart the struct/class to actually apply the attribute to the member variables so emits an error.

This change removes the attributes so we can have the same registerization behaviour that we got in previous releases.

## External Dependencies
 
* N/A

## Type of change

## How Has This Been Tested?

- [X] Command Line
- [ ] oneapi-cli
- [ ] Visual Studio
- [ ] Eclipse IDE
- [ ] VSCode


yuguen-intel (approved):

yuguen-intel:
Looks good to me, waiting for @whitepau review

PR #2460, opened 2024-08-19
Add missing headers to introduce math functions needed by sample.

timmiesmith:
# Existing Sample Changes
## Description

A recent change in the DPC++ compiler has removed an implicit inclusion of cmath and complex header files.  This change adds those explicitly to the same to resolve compile errors that otherwise are encountered with the new DPC++ compiler.

Fixes Issue# 

## External Dependencies

None

## Type of change

- [ X ] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

Confirmed the sample would not compile on my local system as-is, and that adding the header files as suggested resolved the issue.

- [ ] Command Line
- [X ] oneapi-cli
- [ ] Visual Studio
- [ ] Eclipse IDE
- [ ] VSCode

### DIFF
--- a/DirectProgramming/C++SYCL/ParallelPatterns/PrefixSum/src/PrefixSum.cpp
+++ b/DirectProgramming/C++SYCL/ParallelPatterns/PrefixSum/src/PrefixSum.cpp
@@ -34,6 +34,8 @@
 
 #include <iostream>
 #include <string>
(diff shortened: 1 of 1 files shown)


timmiesmith:
Fix has been applied to development branch already.

PR #2502, opened 2024-09-27
FPGA: Update README to include note about emulator environment variables

justin-rosner:
## Description

The FPGA Emulator has modified how pipe depth is emulated. If a design wishes to increase the host buffer capacity (in emulation) of a host pipe to larger than the buffer on the device, `CL_CONFIG_CHANNEL_DEPTH_EMULATION_MODE` will have to be set to `ignore-depth`.

This change updates the README's of designs that do exactly that, and were previously relying on the emulator defaulting to this behaviour.

## External Dependencies

* N/A

## Type of change

## How Has This Been Tested?

- These are just changes to the README. No code was touched.

### DIFF
--- a/DirectProgramming/C++SYCL_FPGA/ReferenceDesigns/convolution2d/README.md
+++ b/DirectProgramming/C++SYCL_FPGA/ReferenceDesigns/convolution2d/README.md
@@ -363,6 +363,8 @@ This design uses CMake to generate a build script for GNU/make.
    | FPGA Simulator      | `make fpga_sim`
    | FPGA Hardware       | `nmake fpga`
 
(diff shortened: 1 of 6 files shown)


whitepau:
@justin-rosner I thought we were setting the 'infinite host pipe host buffer depth' behaviour to be default:

https://hsdes.intel.com/appstore/article/#/14019216094

why do we need to set an environment variable?

justin-rosner:
> @justin-rosner I thought we were setting the 'infinite host pipe host buffer depth' behaviour to be default:
> 
> https://hsdes.intel.com/appstore/article/#/14019216094
> 
> why do we need to set an environment variable?

We made that change for the simulator, but when testing the latest sycl/rel resource it appears that the emulator team made the opposite decision. We will have to file a case to the emulator team for the 2025.1 release to have the default once again be infinite pipe depth .

justin-rosner:
The change that caused this behaviour was reverted, so we no longer need this change.

PR #2561, opened 2024-12-18
[DeviceASAN] Remove not meaningful samples and update license

AllanZyne:
Some asan samples are not related to DenseLinearAlgebra and pure functional tests. So I suggest to remove them.


AllanZyne:
Hi @jimmytwei, could you help review this PR? thanks!

jimmytwei:
@IgorOchocki  Can you take a look at this PR and review it?

IgorOchocki:
So this sample is all about error control and we don't have a proper directory for this. I think we can create a folder called ErrorControl and move the asan sample there.

AllanZyne:
> So this sample is all about error control and we don't have a proper directory for this. I think we can create a folder called ErrorControl and move the asan sample there.

If so, I think a folder named "Sanitizer" is also good enough.
If your team agree, I can help create one in this PR.
But I still need to modify or remove some of tests, because I think some of them isn't proper to be sample codes to users.

AllanZyne:
I need to fix tc test ASAP, so let's start from a simple solution: https://github.com/oneapi-src/oneAPI-samples/pull/2562

PR #2591, opened 2025-03-03
Update name and old links according to AI Tools

bopeng1234:
## Description

Update name and old links according to AI Tools in Readme


Ankur-singh:
@ishaghosh27 can you please have a look at this PR? Thank you 😁

jimmytwei (approved):
@Ankur-singh I think this is good. We should be using AI Tools instead of AI Analytics Toolkit. We'll have to remove references to the offline installer but that can be in a later PR.

@bopeng1234 Thanks so much for updating all of the samples!

